PR #17, opened 2022-04-09
DB: Setup flow pattern functions and runs read fucntions

benedeki:
* Setup function for flow patterns
* Some selected read functions for runs - both segmentation and flow based.

### DIFF
--- a/database/src/main/flow_patterns/insert_additional_data.sql
+++ b/database/src/main/flow_patterns/insert_additional_data.sql
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


database/src/main/flow_patterns/insert_additional_data.sql
```diff
@@ -0,0 +1,75 @@
...
+--
+-- Returns:
+--      status                  - Status code
+--      status_text             - Status text
+--      id_fp_additional_data   - id of the additional data pattern entry
+--      key_fp_flow             - id of the flow the additional data pattern were added into
+--
+-- Status codes:
+--      11                  - Additional data entry created
+--      41                  - Flow not found
```
dk1844, line 47: Is there a reason why this is `41 Flow not found` (as in type `41 - master record not found`), while other calls respond by `40 Flow not found` (as in type `40 - general not found`)?

dk1844 (approved):
Except for seemingly 40/41 error codes being both used for apparently a similar situation (please correct me if I am wrong), I have not found anything suspicious (LGTM, just read the code)

database/src/main/flow_patterns/insert_additional_data.sql
benedeki: In this case master record is not found in an attempt to add data from master-detail relationship.
But it's totally possible to use 40 as well, if the contact between DB and app would consider in more useful 😉

benedeki: In this case master record is not found in an attempt to add data from master-detail relationship.
But it's totally possible to use 40 as well, if the contact between DB and app would consider in more useful 😉

PR #19, opened 2022-04-11
#16: Implement 'runs.get_flow_additional_data` db function

TerezaHrubesova:


database/src/main/runs/get_flow_additional_data.sql
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0 *
```
Zejnilovic, line 7:
```suggestion
 *     http://www.apache.org/licenses/LICENSE-2.0 
 *
```

database/src/main/runs/get_flow_additional_data.sql
```diff
@@ -0,0 +1,85 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION runs.get_flow_additional_data(
+    IN  i_id_flow             BIGINT,
+    OUT status                INTEGER,
+    OUT status_text           TEXT,
+    OUT flow_name             TEXT,
+    OUT id_segmentation       BIGINT,
```
benedeki, line 20: I would perhaps leave these two fields out.

database/src/main/runs/get_flow_additional_data.sql
```diff
@@ -0,0 +1,85 @@
...
+--      10                  - OK
+--      41                  - Flow not found
+--
+-------------------------------------------------------------------------------
+DECLARE
+    _flow_name   TEXT;
+BEGIN
+    _flow_name = (SELECT flow_name FROM runs.flows WHERE id_flow = i_id_flow);
+
+    IF _flow_name IS NULL THEN
```
benedeki, line 64:
```suggestion
    SELECT flow_name 
    FROM runs.flows 
    WHERE id_flow = i_id_flow
    INTO _flow_name;

    IF NOT found THEN
```
I would suggest this style. The other is not particularly wrong, but this makes it somehow clearer, and also the `found` shows that the query cannot return row with `NULL` in `flow_name`.

database/src/main/runs/get_flow_additional_data.sql
```diff
@@ -0,0 +1,85 @@
...
+    SELECT 10, 'OK', _flow_name AS flow_name, sgm.id_segmentation, sgm.segmentation, ad.id_additional_data,
+    ad.ad_name, ad.ad_value, ad.created_by, ad.created_at, ad.updated_by, ad.updated_at
+    FROM runs.segmentation_to_flow sgmtf
+    INNER JOIN runs.segmentations sgm
+      ON sgmtf.key_segmentation = sgm.id_segmentation
+    INNER JOIN runs.additional_data ad
+      ON sgm.id_segmentation = ad.key_segmentation
+    WHERE sgmtf.key_flow = i_id_flow;
+
+END;
```
benedeki, line 81:
```suggestion
    RETURN;
END;
```

benedeki (changes requested):
Ditto the other function. Great work! 😎

database/src/main/runs/get_flow_additional_data.sql
TerezaHrubesova: Removed.

database/src/main/runs/get_flow_additional_data.sql
TerezaHrubesova: Removed.

database/src/main/runs/get_flow_additional_data.sql
TerezaHrubesova: Changed.

database/src/main/runs/get_flow_additional_data.sql
Zejnilovic: https://www.apache.org/licenses/LICENSE-2.0#apply

To be perfectly precise. There should be a new line between the link as outlined in the `apply` section of the license

@AB006J9 applies to your PR as well.

database/src/main/runs/get_flow_additional_data.sql
```diff
@@ -0,0 +1,86 @@
...
+--
+-------------------------------------------------------------------------------
+DECLARE
+    _flow_name   TEXT;
+BEGIN
+
+    SELECT flow_name
+    FROM runs.flows
+    WHERE id_flow = i_id_flow
+    INTO _flow_name;
```
benedeki, line 62: A lesson not to listen to my suggestion literally. I suggested this, while few lines above I suggested to ditch the `flow_name` completely. So more efficient is:
```suggestion
    PERFORM 1
    FROM runs.flows
    WHERE id_flow = i_id_flow;
```
Of course the local `_flow_name` is not needed anymore.

database/src/main/runs/get_flow_additional_data.sql
TerezaHrubesova: I see. New line added.

database/src/main/runs/get_flow_additional_data.sql
TerezaHrubesova: Updated.

benedeki (dismissed):
Similarly to the other, pleas don't merge yet.

dk1844 (dismissed):
LGTM (just read the code)

benedeki (approved):

Issue #14, opened 2022-04-08
Implement `runs.get_segmentation_measure_definitions` db function


benedeki:
Create a DB function that returns all the checkpoint measure definitions bound to the provided segmentation.

The function should return a set of entries, basically the columns of additional data table and status.

Base it on [/feature/first-draft-of-db-data-model-v2](https://github.com/AbsaOSS/atum-service/tree/feature/first-draft-of-db-data-model-v2) branch.

Take [get_segmentation_additional_data.sql](https://github.com/AbsaOSS/atum-service/blob/feature/db-read-and-setup-functions/database/src/main/runs/get_segmentation_additional_data.sql) as example.

PR #22, opened 2022-04-29
#14 implement runs.get segmentation measure definitions db function

AB006J9:


dk1844 (approved):
LGTM from the structure of PR PoV, otherwise OK. (just read the code)

PR #42, opened 2023-06-09
34 transfer control measure atum

### LINKED ISSUE #34: Transfer control measure abilities from Atum library
## Background
[Atum library](https://github.com/AbsaOSS/atum) already implemented the routines to measure a number of control measure types on data in `DataFrame`

## Feature
Copy the code from [Atum library](https://github.com/AbsaOSS/atum) that computes the control measures on `DataFrame` data.

seldon851:
The scope of this PR is to add calculation capabilities to the new Atum module

Closes #34


agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  def context(measurement: Iterable[Measurement])(implicit
+    atumAgent: AtumAgent
+  ): AtumContext =
+    AtumContext(measurements = measurement.map(m => m.name -> m).toMap)
+
+  def context(measurement: Measurement)(implicit
+    atumAgent: AtumAgent
+  ): AtumContext =
+    AtumContext(measurements = Map(measurement.name -> measurement))
```
benedeki, line 92: I don't see the sense in these functions.

agent/src/main/scala/za/co/absa/atum/agent/core/MeasurementProcessor.scala
```diff
@@ -0,0 +1,105 @@
...
+  type MeasurementFunction = DataFrame => String
+
+  def getFunction(
+    dataset: Dataset[Row],
+    measurement: Measurement
+  ): MeasurementFunction
+
+}
+
+class MeasurementProcessorImplementation extends MeasurementProcessor {
```
benedeki, line 20: Instead of this implementation I would - with the companion object - define classes inheriting from `MeasurementProcessor`, one for  each measurement class.
If there would be an enumeration for measurement types, this would ensure at compile time that for each `Measurement` exists a `MeasurementProcessor`.

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -0,0 +1,22 @@
...
+/**
+ *  Place holder for the agent that communicate with the API.
+ */
+trait AtumAgent {
+
+  def measurePublish(measure: Measurement): Unit
+}
+
+class AtumAgentImpl extends AtumAgent {
+  override def measurePublish(measure: Measurement): Unit =
```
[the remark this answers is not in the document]
seldon851: Good idea, we can have both. Accept a single measure or accept an `AtumContext` and publish all.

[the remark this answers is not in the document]
seldon851: I think this should accept both possibilities. The library accept both, working with a Context or sending simple measures, let's talk about this in our meeting

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,94 @@
...
+      df
+    }
+
+    /**
+     *  Set a point in the pipeline to execute calculation.
+     *  @param atumContext Contains the calculations to be done and publish the result
+     *  @return
+     */
+    def setCheckpoint(
+      atumContext: AtumContext
```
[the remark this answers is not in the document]
seldon851: Changing the name, but what if we got multiple contexts in the same scope?

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -0,0 +1,22 @@
...
+    println("Enqueued measurement: " + (measure))
+
+}
+
+object AtumAgentImpl {
+
+  def agent = new AtumAgentImpl
+}
```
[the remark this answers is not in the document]
seldon851: Doing in the next commit

agent/src/main/scala/za/co/absa/atum/agent/core/MeasurementProcessor.scala
```diff
@@ -0,0 +1,105 @@
...
+      case SumOfValuesOfColumn(_, controlCol, _) =>
+        (ds: Dataset[Row]) => ds.select(col(controlCol)).count().toString
+      case AbsSumOfValuesOfColumn(_, controlCol, _) =>
+        (ds: Dataset[Row]) =>
+          val aggCol = sum(abs(col(valueColumnName)))
+          aggregateColumn(ds, controlCol, aggCol)
+
+      case SumOfHashesOfColumn(_, controlCol, _) =>
+        (ds: Dataset[Row]) =>
+          val aggColName = ControlMeasureUtils.getTemporaryColumnName(ds)
```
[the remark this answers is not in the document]
seldon851: done in the next commit

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import org.apache.spark.sql.DataFrame
+import za.co.absa.atum.agent.model.{MeasureResult, Measurement}
+
+/**
+ *  AtumContext: This class provides the methods to measure Spark `Dataframe`. Also allows to add/edit/remove measures.
+ *  @param measurements: A sequences of measurements.
+ */
+
+case class AtumContext(measurements: Map[String, Measurement] = Map()) {
```
benedeki, line 11: Measurement don't (shouldn't) have name. Perhaps `Set` is enough. Or `Seq` if order matters, but IMHO it doesn't.

agent/src/main/scala/za/co/absa/atum/agent/core/MeasurementProcessor.scala
```diff
@@ -0,0 +1,13 @@
+package za.co.absa.atum.agent.core
+
+import org.apache.spark.sql.DataFrame
+
+trait MeasurementProcessor {
+
+  type MeasurementFunction = DataFrame => String
```
benedeki, line 7: Would put it into the companion object. Not to face path-dependent type limitations

agent/src/main/scala/za/co/absa/atum/agent/core/MeasurementProcessor.scala
```diff
@@ -0,0 +1,13 @@
...
+
+  def getMeasureFunction: MeasurementFunction
+
+  def f: MeasurementFunction = getMeasureFunction
```
benedeki, line 11: Why having both?
Wouldn't actually the `apply` function work here the best?

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.spark.sql.types.{DecimalType, LongType, StringType}
+import za.co.absa.atum.agent.core.MeasurementProcessor
+import org.apache.spark.sql.functions._
+import za.co.absa.spark.commons.implicits.StructTypeImplicits.StructTypeEnhancements
+
+/**
+ *  Type of different measures to be applied to the columns.
+ */
+trait Measurement extends MeasurementProcessor {
+  val name: String
```
benedeki, line 14: `name` not needed.

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -0,0 +1,137 @@
...
+
+    override def getMeasureFunction: MeasurementFunction =
+      (ds: Dataset[Row]) => ds.select(col(controlCol)).count().toString
+
+  }
+
+  case class DistinctRecordCount(
+    name: String,
+    controlCol: String,
+    resultValue: Option[String] = None
```
benedeki, line 33: `resultValue` should be removed. (multiple cases)

agent/src/test/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  @transient var ss: SparkSession = null
+  @transient var sc: SparkContext = null
+
+  private object testImplicits extends SQLImplicits {
+    protected override def _sqlContext: SQLContext = self.ss.sqlContext
+  }
+
+  override def beforeAll(): Unit = {
+    val sparkConfig = new SparkConf()
```
benedeki, line 22: You can use `SparkTestBase` from [spark-commons](https://github.com/AbsaOSS/spark-commons). It provides spark session in `spark` variable.

benedeki:
Added _Closes #34_ line to description to link the GitHub PR ticket to the GitHub issue.

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,66 @@
...
+      measures.foreach(m => executeMeasure(checkpointName, m))
+      df
+    }
+
+    /**
+     *  Set a point in the pipeline to execute calculation.
+     *  @param atumContext Contains the calculations to be done and publish the result
+     *  @return
+     */
+    def createCheckpoint(checkpointName: String, atumContext: AtumContext): DataFrame = {
```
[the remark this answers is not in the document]
seldon851: @benedeki David, it is foreseen that we can have more than one `AtumContext` in the same scope.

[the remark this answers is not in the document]
seldon851: it is foreseen that we can have more than one `AtumContext` in the same scope.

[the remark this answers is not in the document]
seldon851: changed

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
seldon851: removed

.scalafmt.conf
```diff
@@ -0,0 +1,21 @@
+version = "3.5.3"
+runner.dialect = scala211
```
lsulak, line 2: 2.11 or 2.12 ? Project is gonna support both, thus you chosen the smaller one out of them..?

build.sbt
```diff
@@ -15,8 +15,7 @@
 
 import Dependencies._
 
-
-ThisBuild / organization     := "za.co.absa"
+ThisBuild / organization := "za.co.absa"
 
 lazy val scala211 = "2.11.12"
```
lsulak, line 20: This is never used, right?

project/Dependencies.scala
```diff
@@ -13,11 +13,10 @@
  * limitations under the License.
  */
 
-import sbt._
 
-object Dependencies {
+import sbt.*
```
lsulak, line 17:
```suggestion
import sbt._
```

❓

project/Dependencies.scala
```diff
@@ -54,7 +53,32 @@ object Dependencies {
...
+    val spark3Version = "3.3.2"
+    val scala212 = "2.12.12"
+    val scalatestVersion = "3.2.15"
+    val specs2Version = "4.19.2"
+    val typesafeConfigVersion = "1.4.2"
+
+    lazy val sparkCore = "org.apache.spark" %% "spark-core" % spark3Version /*% Provided*/ /*exclude(
+        "com.fasterxml.jackson.core", "jackson-databind"
+      ) exclude(
+        "com.fasterxml.jackson.module", "jackson-module-scala_" + scala212.substring(0, 4) // e.g. 2.11
```
lsulak, line 68: dead code, should be removed?

README.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Modules
+
+### Agent `agent/`
+This module is intended to replace the current [Atum](https://github.com/AbsaOSS/atum) repository. Provides libraries for establishing and pushing them to the API located in `server/`.
+See `agent/README.md`.
+
+### Server `server/`
+An API under construction that communicates with AtumAgent and with the persisting storage. It also provides measure configuration to the `AtumAgent`.
+See `server/README.md`.
```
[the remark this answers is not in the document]
seldon851: thank you

.scalafmt.conf
seldon851: let's change it to `runner.dialect = scala212`

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,77 @@
...
+  def withMeasureRemoved(measurement: Measurement): AtumContext =
+    this.copy(measurements = measurements.filterNot(_ == measurement))
+
+}
+
+object AtumContext {
+  implicit class DatasetWrapper(df: DataFrame) {
+
+    /**
+     *  Executes the measure directly with not AtumContext.
```
lsulak, line 42: Executes the measure directly with not AtumContext.

What do you mean? Executes the measure directly outside AtumContext. ? Or when AtumContext not provided?

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,77 @@
...
+ *  AtumContext: This class provides the methods to measure Spark `Dataframe`. Also allows to add/edit/remove measures.
+ *  @param measurements: A sequences of measurements.
+ */
+
+case class AtumContext(measurements: Set[Measurement] = Set()) {
+
+  def withMeasures(
+    measure: Measurement
+  ): AtumContext =
+    this.copy(measurements = Set(measure))
```
lsulak, line 16: so this is more like 'withMeasureReplaced' ?

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def withMeasures(
+    measures: Iterable[Measurement]
+  ): AtumContext =
+    this.copy(measurements = measures.toSet)
```
lsulak, line 21: so this is more like 'withMeasuresReplaced' ?

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def withMeasuresAdded(
+    measure: Measurement
+  ): AtumContext =
+    this.copy(measurements = measurements + measure)
```
lsulak, line 26: so this is more like 'withMeasureAdded' ?

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     *  Set a point in the pipeline to execute calculation.
+     *  @param checkpointName The key assigned to this checkpoint
+     *  @param atumContext Contains the calculations to be done and publish the result
+     *  @return
+     */
+    def createCheckpoint(checkpointName: String)(implicit atumContext: AtumContext): DataFrame = {
+      atumContext.measurements.map { measure =>
```
lsulak, line 65: the result of this map is not stored anywhere, so wouldn't `foreach` be better?

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,77 @@
...
+        val result = MeasureResult(measure, measure.function(df))
+        AtumAgent.publish(checkpointName, atumContext, result)
+
+        executeMeasures(checkpointName, atumContext.measurements)
```
lsulak, line 69: This make more sense, doesn't it? Since you would be executing all the measures for each measure, since the map is kind of a for/finite loop

```suggestion
        executeMeasure(checkpointName, measurement)
```

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     *  Executes the measure directly with not AtumContext.
+     *  @param measure the measure to be calculated
+     *  @return
+     */
+    def executeMeasure(checkpointName: String, measure: Measurement): DataFrame = {
+
+      val result = MeasureResult(measure, measure.function(df))
+      AtumAgent.measurePublish(checkpointName, result)
```
lsulak, line 49: actually, what's the difference between publish and measurePublish?

agent/src/main/scala/za/co/absa/atum/agent/model/MeasureResult.scala
```diff
@@ -0,0 +1,3 @@
+package za.co.absa.atum.agent.model
+
+case class MeasureResult(measurement: Measurement, result: String)
```
lsulak, line 3: I know that this is really just a first PoC / draft, but I'm still interested in your thoughts here - why String? just curious really :)

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -0,0 +1,130 @@
+package za.co.absa.atum.agent.model
+
+import org.apache.spark.sql.{Column, DataFrame, Dataset, Row}
```
lsulak, line 3: DataFrame not used

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.spark.sql.functions.col
+import org.apache.spark.sql.types.{DecimalType, LongType, StringType}
+import za.co.absa.atum.agent.core.MeasurementProcessor
+import org.apache.spark.sql.functions._
+import za.co.absa.atum.agent.core.MeasurementProcessor.MeasurementFunction
+import za.co.absa.atum.agent.model.Measurement.RecordCount
```
lsulak, line 9: No need to import it, it's in this module after all. Maybe this is a refactoring thing, we all are doing these from time to time :D

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -0,0 +1,130 @@
...
+}
+
+object Measurement {
+
+  case class RecordCount(
+    controlCol: String
+  ) extends Measurement {
+
+    override def function: MeasurementFunction =
+      (ds: Dataset[Row]) => ds.select(col(controlCol)).count().toString
```
lsulak, line 27: `Dataset[Row]` is the same as `DataFrame` and you are using `DataFrame` elsewhere, so why not to make it consistent?

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -0,0 +1,130 @@
...
+      val aggregatedColumnName = ds.schema.getClosestUniqueName("sum_of_hashes")
+      val value = ds
+        .withColumn(aggregatedColumnName, crc32(col(controlCol).cast("String")))
+        .agg(sum(col(aggregatedColumnName)))
+        .collect()(0)(0)
+      if (value == null) "" else value.toString
+    }
+  }
+
+  private val valueColumnName: String = "value"
```
lsulak, line 74: I would move this val much more up, maybe even on the beginning of this object

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -0,0 +1,130 @@
...
+      case _ =>
+        val ds2 = ds.select(col(measureColumn).as(valueColumnName))
+        val collected = ds2.agg(aggExpression).collect()(0)(0)
+        if (collected == null) 0 else collected
+    }
+    // check if total is required to be presented as larger type - big decimal
+    workaroundBigDecimalIssues(aggregatedValue)
+  }
+
+  private def workaroundBigDecimalIssues(value: Any): String =
```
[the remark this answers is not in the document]
lsulak: Or maybe let's somehow try to restrict the Any. It smells like Python :D

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
seldon851: In simple pipelines, you can execute many Measures without creating an `AtumContext`.

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
seldon851: yes, but also it creates the measures in case the measures don't exists

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
seldon851: yes but I think that when you are coding is more comfortable using the ide if the method is overloaded

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
seldon851: It is, indeed. I have changed the code and this remained, my fault

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
seldon851: This is a call. One `AtumContext` instance have multiples `Measurement`

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
seldon851: `publish` is the main case. You submit an `AtumContext` and the method sends the data to the API. `measurePublish` is a corner case where you only submit a bunch of `Measurement`.

`AtumContext` is going to provide metadata with `Segmentation` in the next iteration of code so it makes sense to have an AtumContext

agent/src/main/scala/za/co/absa/atum/agent/model/MeasureResult.scala
seldon851: Yes, it's because I took the calculations from the current Atum repo, and all the results are `String`s. I may change this and use a `JSON` instead in the next ticket

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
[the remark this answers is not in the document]
seldon851: we can integrate the `match` into the other method

benedeki (approved):
I think it looks good now.
Some adjustments probably will be added. But let's enable you for the last part while we still have you available 😉

PR #72, opened 2023-09-07
#37: dao between agent server adjustments

### LINKED ISSUE #37: Create/Adjust DAO to use between Agent and Server
## Background
Data needs to be exchanged between the Agent and the Server

## Feature
Create and update  classes for the data exchange so there's no misalignment between the data sent and received.

## Proposed Solution
1. Probably should be a separate module within the Atum project 

Depends on #39

lsulak:
This PR includes:
* Renamed Segmentation -> Partition
* Fixed, refactored, and improved the build and dependency management, almost ready for Scala 2.13 even
* Kept the old model in server part intact, because it would require a lot of changes to change all the endpoints etc
* Created a bare minimum module for data model, for transferring data between agent and server 

Closes #37


model/src/main/scala/za/co/absa/atum/model/AdditionalData.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+case class AdditionalData(
+    additionalData: Map[String, String]
+)
```
cerveada, line 21: Do we need the class if it has just one field? We could use the field directly. The same thing with partitioning.

lsulak: I was thinking the same. I think that potentially if we decide to extend them, which might be highly likely, then it makes sense to keep it as is.

In any case, especially with Partitioning, I wanted to keep the information about the overall partitioning structure close and encapsulated, i.e. as a separated class. If I keep only Partition class and move the `Seq[Partition]` into `Checkpoint`, then I have to implement `Seq[Partition]` on other places just to enforce the order of partitions. By having it like this, everyone can use Partitioning with pre-described overall structure, i.e. saying that Partitions are related to each other in Seq way and order is important.

Not sure if my explanation makes sense though :D ?

@benedeki do you have an opinion about this? What about some sort of partition / metadata versioning? Or some other attributes?

model/src/main/scala/za/co/absa/atum/model/AtumContext.scala
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+case class AtumContext(
+    partitions: Partitioning,
+    measures: Set[Measure] = Set.empty
+)
```
benedeki, line 22: **Less important:**
Missing `additionalInfo`

model/src/main/scala/za/co/absa/atum/model/Checkpoint.scala
```diff
@@ -0,0 +1,27 @@
...
+package za.co.absa.atum.model
+
+import java.time.ZonedDateTime
+
+case class Checkpoint(
+    name: String,
+    partitioning: Partitioning,
+    processStartTime: ZonedDateTime,
+    processEndTime: ZonedDateTime,
+    measurements: Seq[Measurement]
```
benedeki, line 26:
```suggestion
    measurements: Seq[Measurement],
    author: String
```
We need to add the support for "authoring" into `AtumAgent` and `AtumContext` methods.

model/src/main/scala/za/co/absa/atum/model/Measure.scala
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+case class Measure(
+    measureName: String,
```
benedeki, line 20: Name not needed. It's a heritage of original _Atum, proved not to bring value.

model/src/main/scala/za/co/absa/atum/model/Measure.scala
```diff
@@ -0,0 +1,23 @@
...
+    measureFunction: String,
```
benedeki, line 21: `measureFunction` or `measureType`? 🤔

model/src/main/scala/za/co/absa/atum/model/AdditionalData.scala
benedeki: I think at the beginning we can omit these very simple types. (This and for example `Partition`)

benedeki (commented):
Won't it be confusing having same type names in model and agent/server?
Wouldn't it be clearer to name them with suffix DTO?

model/src/main/scala/za/co/absa/atum/model/Measure.scala
lsulak: okay, will change

model/src/main/scala/za/co/absa/atum/model/Measure.scala
lsulak: agreed on `measureFunctionName`

model/src/main/scala/za/co/absa/atum/model/Checkpoint.scala
lsulak: okay

model/src/main/scala/za/co/absa/atum/model/AtumContext.scala
lsulak: okay. Will be empty by default, as we agreed on

cerveada:
Just notice some case classes are indented by 4 spaces, some 2. All should be 2 right?

model/src/main/scala/za/co/absa/atum/model/CheckpointDTO.scala
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+import java.time.ZonedDateTime
+
+case class CheckpointDTO(
```
benedeki, line 21: Per our discussion we need a UUID unique for this checkpoint (generated by the Agent probably)

model/src/main/scala/za/co/absa/atum/model/AtumContextDTO.scala
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+case class AtumContextDTO(
+    partitions: PartitioningDTO,
```
benedeki, line 20: Genuine question, but would like to have the same name both in DTO and DB.
Do we want `partitions` or `partitioning`? 🤔

model/src/main/scala/za/co/absa/atum/model/MeasurementDTO.scala
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+case class MeasurementDTO(
+    measure: MeasureDTO,
+    measureResult: String
```
benedeki, line 21: DB is already build for value types. Is it an extra effort to do it here too? 🤔

model/src/main/scala/za/co/absa/atum/model/CheckpointDTO.scala
lsulak: Yes, will add it

model/src/main/scala/za/co/absa/atum/model/AtumContextDTO.scala
lsulak: todo rename to `partitioning`

lsulak:
> Just notice some case classes are indented by 4 spaces, some 2. All should be 2 right?

you are right. I think that I liked 4 more for the case classes in model/ but I'll change it :D

model/src/main/scala/za/co/absa/atum/model/MeasurementDTO.scala
lsulak: note to myself - this should be the only outstanding issue in this PR for now, otherwise feel free to review again

benedeki:
Ještě bych změnil AdditionalData na Map[String, Option[String]]

model/src/main/scala/za/co/absa/atum/model/dto/CheckpointDTO.scala
```diff
@@ -14,15 +14,14 @@
...
-    partitioning: PartitioningDTO,
-    processStartTime: ZonedDateTime,
-    processEndTime: ZonedDateTime,
-    measurements: Seq[MeasurementDTO]
+  id: UUID,
+  name: String,
+  author: String,
+  partitioning: Partitioning,
+  processStartTime: ZonedDateTime,
+  processEndTime: ZonedDateTime,
```
benedeki, line 25: Sorry, for another late remark. For eventual streaming, this needs to be 
```suggestion
  processEndTime: Option[ZonedDateTime],
```

model/src/main/scala/za/co/absa/atum/model/MeasureResult.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+trait MeasureResult[T]
```
benedeki, line 19: Wonder how this will work for de-/serialization 🤔 😉

benedeki (dismissed):
The comment suggested can be done in a subsequent PR

model/src/main/scala/za/co/absa/atum/model/dto/CheckpointDTO.scala
lsulak: It makes sense.I'll change it now

model/src/main/scala/za/co/absa/atum/model/dto/CheckpointDTO.scala
```diff
@@ -0,0 +1,33 @@
...
+import za.co.absa.atum.model._
+
+import java.time.ZonedDateTime
+import java.util.UUID
+
+case class CheckpointDTO(
+  id: UUID,
+  name: String,
+  author: String,
+  measurementsProvidedManually: Boolean = false,
```
benedeki, line 28: Feels like a mouth-full sorry. 😊 
```suggestion
  measuredByAtum: Boolean = true,
```
(Although I would prefer the default to be `false` 🤔 )

lsulak: how about `measuredByAtumAgent` ? 

Okay, `false` is good, agreed - at least we will explicitly set this to `true` in the Agent and whoever will send measurements that were computed by some other process other than the agent, might not want to care about this parameter

model/src/main/scala/za/co/absa/atum/model/MeasureResult.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+trait MeasureResultBase[T] extends Any { def result: T }
+
+case class MeasureResult[T](result: T) extends AnyVal with MeasureResultBase[T]
```
benedeki, line 21: Probably will change somewhat to enable more fields (case classes cannot be inherited), but OK for now.

model/src/main/scala/za/co/absa/atum/model/Measurement.scala
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.model
+
+case class Measurement(
+  measure: Measure,
+  result: MeasureResult[String]
```
benedeki, line 21: Again, I see this can be perhaps type parametrized eventually too. But as above, now good enough.

benedeki (approved):
Good, actually very good basis. 
While I expect some further evolution before release, let's have this one in

A side note:
The formatting changes on untouched files is... adding noise. 😉

PR #78, opened 2023-09-22
#77: scala2.13 support

### LINKED ISSUE #77: Adding Scala 2.13 into supported versions
## Background
As the Issue description says.

lsulak:
Closes #77


cerveada (dismissed):

build.sbt
```diff
@@ -68,6 +68,15 @@ lazy val server = (projectMatrix in file("server"))
       libraryDependencies ++= Dependencies.serverDependencies,
       Compile / packageBin / publishArtifact := false,
       (Compile / compile) := ((Compile / compile) dependsOn printSparkScalaVersion).value,
+      Compile / unmanagedSourceDirectories += {
```
benedeki, line 71: Server is not a lib to include but a running app. I would simply build it with Scala 2.13 atm

benedeki: Actually no. Keep it only 2.12. fa-db doesn't support 2.13 yet. :-(

lsulak: You are right - I've changed a thing or two. Now only the agent (and model) is being compiled against all 3 versions, the server is compiled only for 2.12.

The sbt log looks something like this:

```
[info] compiling 7 Scala sources to /Users/ab0222l/Documents/git/atum-service/model/target/jvm-2.12/classes ...
[info] compiling 7 Scala sources to /Users/ab0222l/Documents/git/atum-service/model/target/jvm-2.13/classes ...
[info] compiling 7 Scala sources to /Users/ab0222l/Documents/git/atum-service/model/target/jvm-2.11/classes ...
[info] compiling 25 Scala sources to /Users/ab0222l/Documents/git/atum-service/server/target/jvm-2.12/classes ...
[info] compiling 5 Scala sources to /Users/ab0222l/Documents/git/atum-service/agent/target/spark3-jvm-2.12/classes ...
[info] compiling 5 Scala sources to /Users/ab0222l/Documents/git/atum-service/agent/target/jvm-2.12/classes ...
[info] compiling 5 Scala sources to /Users/ab0222l/Documents/git/atum-service/agent/target/jvm-2.13/classes ...
[info] compiling 5 Scala sources to /Users/ab0222l/Documents/git/atum-service/agent/target/spark2-jvm-2.12/classes ...
[info] compiling 5 Scala sources to /Users/ab0222l/Documents/git/atum-service/agent/target/spark3-jvm-2.13/classes ...
[info] compiling 5 Scala sources to /Users/ab0222l/Documents/git/atum-service/agent/target/jvm-2.11/classes ...
[info] compiling 5 Scala sources to /Users/ab0222l/Documents/git/atum-service/agent/target/spark2-jvm-2.11/classes ...
```

Zejnilovic:
https://github.com/AbsaOSS/atum-service/blob/cdf18c2c5073b4b339178aa0ccfa8cb20844a93d/.github/workflows/jacoco_check.yml#L36C27-L36C27

Please update this to from `2.12.12` to `2.12`

miroslavpojer:
> https://github.com/AbsaOSS/atum-service/blob/cdf18c2c5073b4b339178aa0ccfa8cb20844a93d/.github/workflows/jacoco_check.yml#L36C27-L36C27
> 
> Please update this to from `2.12.12` to `2.12`

I have updated it to 2.12.18. That fixed missing JaCoCo comments.

project/Dependencies.scala
```diff
@@ -23,8 +23,11 @@ object Dependencies {
     val spark3 = "3.3.2"
 
     val scala211 = "2.11.12"
-    val scala212 = "2.12.12"
-    val supportedScalaVersions: Seq[String] = Seq(scala211, scala212)
+    val scala212 = "2.12.18"
+    val scala213 = "2.13.11"
+
+    val serverScalaVersion: String = scala212
```
benedeki, line 29: Nice!

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

.github/workflows/jacoco_check.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           # The codebase for all Scala versions is the same, so the coverage is calculated only once
           # Scala 2.12 is chosen since it is supported by the most wide range of Spark versions and
           # vendor distributions.
-          - scala: 2.12.12
+          - scala: 2.12.18
```
Zejnilovic, line 36:
```suggestion
          - scala: 2.12
```

I don't think there is a need for a full scala version here, only the short one is needed and all the long formats can be transformed into short ones.

project/Dependencies.scala
Zejnilovic: Why is this not a Seq as well as `aqentSupportedVersions`?

lsulak: Because supporting more versions in the server is not needed imho, i.e. there will be probably just 1 instance of server, managed by 'us', deployed into Docker & ECR repo & EKS, and it will be able to communicate with the agent that might  be in any version - it's just a REST API communication between them after all

lsulak: but on the other hand, agent will be used inside of our applications' scala code, which are usually in range of Scala 2.11 - 2.13 so that must be a Seq

Zejnilovic: Also, is there a reason for sever not being 2.13?

build.sbt
```diff
@@ -68,6 +68,7 @@ lazy val server = (projectMatrix in file("server"))
       libraryDependencies ++= Dependencies.serverDependencies,
       Compile / packageBin / publishArtifact := false,
       (Compile / compile) := ((Compile / compile) dependsOn printSparkScalaVersion).value,
+      crossScalaVersions := Seq(Versions.serverScalaVersion),
```
Zejnilovic, line 71: What is the purpose of this setting?

.github/workflows/jacoco_check.yml
lsulak: There is `scalaShort` below that we are using for some purposes of having just major+minor versions, but to be fair I don't care that much, whatever works - I borrowed this from https://github.com/AbsaOSS/spark-data-standardization/blob/master/.github/workflows/jacoco_check.yml and checked a few other projects as well, so I haven't really implemented this by myself

build.sbt
```diff
@@ -58,6 +57,7 @@ lazy val root = (projectMatrix in file("."))
     name := "atum-service-root",
     javacOptions ++= Seq("-source", "1.8", "-target", "1.8", "-Xlint"),
     publish / skip := true,
+    crossScalaVersions := Nil,
     mergeStrategy
```
Zejnilovic, line 61: Are we using `sbt assembly` for something?

build.sbt
```diff
@@ -58,6 +57,7 @@ lazy val root = (projectMatrix in file("."))
     name := "atum-service-root",
     javacOptions ++= Seq("-source", "1.8", "-target", "1.8", "-Xlint"),
     publish / skip := true,
+    crossScalaVersions := Nil,
```
Zejnilovic, line 60: What is the purpose of this setting?

build.sbt
```diff
@@ -24,7 +24,6 @@ ThisBuild / organization := "za.co.absa"
 ThisBuild / name         := "atum-service"
```
Zejnilovic, line 24: Please remove the name, not applied

build.sbt
lsulak: hmm, okay I think it was a remnant from the past, will delete it

build.sbt
lsulak: I think I took it from the sbt documentation: https://www.scala-sbt.org/1.x/docs/Cross-Build.html

`crossScalaVersions must be set to Nil on the aggregating project`

so this is Nil here only for the root one

build.sbt
lsulak: I guess that there will be Jar containing server and its configuration, and the model, nicely packaged together with it

build.sbt
lsulak: done

build.sbt
Zejnilovic: But you are using projectmatrix, not the stateful build. So using `crossScalaVersions` makes no sense or am I missing something?

build.sbt
Zejnilovic: I wanna make a bold comment that you do not need `mergeStrategy` at all. Or at least I have not seen it.

project/Dependencies.scala
Zejnilovic: > Because supporting more versions in the server is not needed imho, i.e. there will be probably just 1 instance of server, managed by 'us', deployed into Docker & ECR repo & EKS, and it will be able to communicate with the agent that might be in any version - it's just a REST API communication between them after all

@lsulak  I get it, but then you have 2 different implementations. I would have it both as Seq to have it the same across. First look you see one as Seq and one not.

miroslavpojer:
> https://github.com/AbsaOSS/atum-service/blob/cdf18c2c5073b4b339178aa0ccfa8cb20844a93d/.github/workflows/jacoco_check.yml#L36C27-L36C27
> 
> Please update this to from `2.12.12` to `2.12`

I have tested this "requirement" on another GHA workflow and it does not work. The not shorted scala version string is required in titles (can be changed, here) and in sbt ++{scala.version} call, where it is required.

miroslavpojer (commented):
- [x] pulled
- [x] code review
- [ ] run test
- [x] run jacoco

Optional: This is nice PR to do some atomic increment increase of code codverage.

project/Dependencies.scala
lsulak: Why to have something as a different, and improper, data type, just for the sake of consistency? To the best of my knowledge, server will always be in 1 version, thus any kind of iterable doesn't make sense

lsulak:
Oopsie. We recently added (outside of this PR) logging via SLF4S, but that's not yet compatible with Scala 2.13 :( https://github.com/mattroberts297/slf4s/issues/20

project/Dependencies.scala
benedeki: I like the way how it is.
But the DB tests will also suffice to exist under one Scala version. On the other hand the Model module is multi-versioned. So maybe a rename  of the variable to more neutral ones?

benedeki (approved):

benedeki:
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

Zejnilovic (approved):

benedeki (approved):

PR #79, opened 2023-09-25
#26 add create checkpoint function

### LINKED ISSUE #26: Create functions to create a checkpoint
## Background
For the Atum/Control measure info point of view, it's the key stone to be able to initiate the control measures measurement within a checkpoint of the specified  _Segmentation_.

## Feature
Create a function to execution checkpoint creation for the provides _Segmentation_ on the provided _DataFrame_

Solution Ideas:
1. Create above mentioned function withing the _Segmentation_ class
2. Consider having the checkpoint creation function as an implicit enhancement of `DataFrame`

Depends on #25

cerveada:
Work done mostly on Agent side:
* Lot of refactorization and simplifications
* Adding proper scope (protected / private) of methods and classes
* Finished Checkpoint creation - on Agent as well as Provided data measurements
* Adding a result type for each Measure and implementing validation of this, especially for measurements provided by user
* Slightly changing Agent's internal data model
* Changed `functionName` -> `measureName` to be consistent across the project 
* Lots of unit tests covering all these

Closes #26


agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -39,8 +39,11 @@ class AtumContext private[agent](
     parentAgent.getOrCreateAtumSubContext(atumPartitions ++ subPartitions)(this)
   }
 
-  def createCheckpoint(checkpointName: String, dataToMeasure: DataFrame) = {
-    ??? //TODO #26
+  def createCheckpoint(checkpointName: String, df: DataFrame): Unit = {
+    measures.foreach { measure =>
+      val result = MeasureResult(measure, measure.function(df))
+      AtumAgent.publish(checkpointName, this, result)
```
lsulak, line 45: Hmm so this will basically send many messages to the server, but for batch processing I think that all measurements should be sent in a single message, i.e. compute the results and gather them, and then call `publish`. This would require to change the existing API, or add another definition of the `publish` method, that would have multiple `measureResult` objects on its input.

Streaming will allow partial measurements, i.e. also sending a single measurement could make sense, and the server/DB would connect it together.

What do you think @benedeki ? Also, once my https://github.com/AbsaOSS/atum-service/pull/72 will be merged, you can directly work with `CheckpointDTO` object so this will be changed. But I would still like to talk about the concept of sending a single measurement vs multiple.

cerveada: I agree, I didn't want to guess the DAO format before it's merged, so this is done using the current dispatcher interface. After DAO is merged, I would prefer to change all this in different PR.

lsulak: all right then 👍

TebaleloS: I also agree with @lsulak.

lsulak (approved):

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
benedeki: > I agree, I didn't want to guess the DAO format before it's merged, so this is done using the current dispatcher interface. After DAO is merged, I would prefer to change all this in different PR.

What's the benefit of an extra PR. This one is still tiny. 😉

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -39,8 +39,11 @@ class AtumContext private[agent](
     parentAgent.getOrCreateAtumSubContext(atumPartitions ++ subPartitions)(this)
   }
 
-  def createCheckpoint(checkpointName: String, dataToMeasure: DataFrame) = {
-    ??? //TODO #26
+  def createCheckpoint(checkpointName: String, df: DataFrame): Unit = {
```
benedeki, line 42: Would be returning `AtumContext` (self) to allow chaining.

salamonpavel (dismissed):

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -42,16 +42,42 @@ class AtumContext private[agent] (
...
+      measurements = measurements.toSeq
+    )
   }
 
-  def createCheckpointOnProvidedData(
-                                      checkpointName: String,
-                                      author: String,
-                                      measurements: Seq[Measurement]
-                                    ): Checkpoint = {
+  def createAndSaveCheckpoint(checkpointName: String, author: String, dataToMeasure: DataFrame): AtumContext = {
```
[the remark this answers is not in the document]
lsulak: Okay, will change this consistently on all places & document

agent/src/main/scala/za/co/absa/atum/agent/exception/MeasurementProvidedException.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.agent.exception
+
+case class MeasurementProvidedException(msg: String) extends Exception(msg)
```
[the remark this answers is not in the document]
lsulak: 100%

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -33,85 +34,118 @@ sealed trait Measure extends MeasurementProcessor with MeasureType {
 trait MeasureType {
   val measureName: String
   val onlyForNumeric: Boolean
```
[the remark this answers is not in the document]
lsulak: yes, `avg` can be for non-numeric, so it would be `false, and `sum` probably doesn't make sense so would be `true`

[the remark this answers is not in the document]
lsulak: so basically if it's `false` then no check would be performed, if it's `true` then numerical check would be performed-just a general one basically, but if this is a bad idea I can drop it. I was actually inspired by original Atum library, it had this there in a very similar fashion

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -62,6 +88,16 @@ class AtumContext private[agent] (
     )
   }
 
+  def createAndSaveCheckpointOnProvidedData(
+    checkpointName: String, author: String, measurements: Seq[Measurement]
```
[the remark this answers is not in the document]
lsulak: Agreed to potentially change this once we'll be refactoring the measurement builder & Measure class - all this will be changed to some degree (some changes will be more brutal) in https://github.com/AbsaOSS/atum-service/issues/98

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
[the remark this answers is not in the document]
benedeki: As it's not used, and eventually only part of validation, I would remove it.

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -16,4 +16,62 @@
...
+  }
+}
+
+/**
+ * When the Atum Agent itself performs the measurements, using Spark, then in some cases some adjustments are
+ * needed - thus we are converting the results to strings always - but we need to keep the information about
+ * the actual type as well.
+ */
+case class MeasurementByAtum(measure: Measure, resultValue: String, resultType: ResultValueType.ResultValueType)
+  extends Measurement
```
benedeki, line 77: Do we need these two classes?
If we do, please put them into a companion object to keep the in this file.

lsulak: Yes they are needed; why - what benefit would it have? I think that in this case it would only slightly complicate their usages - if the only reason is encapsulation, then they are, after all, in their own module 🤔

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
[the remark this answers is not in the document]
lsulak: Eh okay, I can remove it here

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
benedeki: I am old school 👼 , don't like top level classes present in a file named by another class.
Also they are `Measurement` descendants, so I see no problem to hide them inside. 
Usage is not an issue with import.

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -43,24 +44,48 @@ class AtumContext private[agent] (
...
   }
 
-  def createCheckpoint(checkpointName: String, author: String, dataToMeasure: DataFrame) = {
-    ??? // TODO #26
+  private def takeMeasurements(df: DataFrame): Set[MeasurementByAtum] = {
+    measures.map { m =>
+      val measurementResult = m.function(df)
+      MeasurementByAtum(m, measurementResult.result, measurementResult.resultType)
+    }
+  }
```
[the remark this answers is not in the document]
benedeki: OK, fair enough.

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

Issue #23, opened 2023-02-08
Create the FA-DB entities for the DB functions


benedeki:
## Background
When the database model is ready the data there should be accessed via the provided database functions.

## Feature
Create the _Fa-db_ classes for the Postgres DB functions Atum database offers.

Depends on #49 
Depends on #69
Depends on #71

benedeki:
This item depends on:
* AbsaOSS/atum-service#49
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

benedeki:
This item depends on:
* ~~AbsaOSS/atum-service#49~~
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

PR #82, opened 2023-09-28
#48: Review of the db model and functions

### LINKED ISSUE #48: Review of the db model and functions
## Background
Models is roughly ready, but changes are to be expected when integration happens and new features emerge.

## Feature
Adjust the db objects according to needs.

benedeki:
* replaced HSTORE with JSONB
* segmentation expression changed to partitioning
* major refactoring of functions and tables
* root files renamed to adhere to fa-db `deploy_db.py` script
* functions not (fully) implemented moved to subfolders `future`

**NB!**
Please ignore the content of the `future` folders.  They contain eventually useful code that has been developed before, but right now out of scope.

Closes #48


database/src/main/01_users.ddl
```diff
@@ -13,26 +13,69 @@
...
+
+
+DO
+$do$
+    BEGIN
+        IF EXISTS (
+                SELECT FROM pg_catalog.pg_roles
+                WHERE  rolname = 'atum_configurator') THEN
+
+            RAISE NOTICE 'Role "atum_configurator" already exists. Skipping.';
```
lsulak, line 69: What it does? I understand the user and owner, I understand the master postgres, but what is the purpose of this guy?

database/src/main/runs/_get_id_partitioning.sql
```diff
@@ -24,22 +24,22 @@ $$
...
+    WHERE PAR.partitioning = i_partitioning
+    INTO id_partitioning;
 
     RETURN;
 END;
 $$
 LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
 
-GRANT EXECUTE ON FUNCTION runs._get_key_segmentation(hstore) TO atum_user;
+ALTER FUNCTION runs._get_id_partitioning(JSONB) OWNER TO atum_owner;
```
lsulak, line 45: why not GRANT EXECUTE and why owner and not the user? I thought that only DB, schemas, and tables are owned by owner, and that the functions are supposed to be executed as the user, not owner

lsulak: or rather, is the idea here that these 'private' functions shouldn't be called by the user at all? Would it work if the user executed a function that calls this function?

database/src/main/runs/future/write_measurement.sql
```diff
@@ -13,13 +13,13 @@
  * limitations under the License.
  */
 
+
 CREATE OR REPLACE FUNCTION runs.write_measurement(
-    IN  i_segmentation          HSTORE,
-    IN  i_id_measurement        UUID,
+    IN  i_segmentation          JSONB,
     IN  i_id_checkpoint         UUID,
```
lsulak, line 19: hmm is `i_id_measurement` not needed here anymore?

lsulak: oh okay I think this function is far from being finished

database/src/main/runs/checkpoint_measure_definitions.ddl
```diff
@@ -13,20 +13,18 @@
...
--- DROP TABLE IF EXISTS runs.checkpoint_measure_definitions;
-
 CREATE TABLE runs.checkpoint_measure_definitions
 (
     id_checkpoint_measure_definition    BIGINT NOT NULL DEFAULT global_id(),
-    key_segmentation                    BIGINT NOT NULL,
-    measure_type                        TEXT NOT NULL,
-    measure_fields                      TEXT[] NOT NULL,
+    fk_partitioning                     BIGINT NOT NULL,
+    function_name                       TEXT NOT NULL,
```
lsulak, line 20: `measure_name` is also good, what do you think?

database/src/main/runs/write_checkpoint.sql
```diff
@@ -0,0 +1,105 @@
...
+    IN  i_control_columns       TEXT[][],
+    IN  i_measure_values        JSONB[],
+    IN  i_by_user               TEXT,
+    OUT status                  INTEGER,
+    OUT status_text             TEXT
+) RETURNS record AS
+$$
+-------------------------------------------------------------------------------
+--
+-- Function: runs.write_measurement(9)
```
lsulak, line 33: todo name

database/src/main/runs/write_checkpoint.sql
```diff
@@ -0,0 +1,105 @@
...
+CREATE OR REPLACE FUNCTION runs.write_checkpoint(
+    IN  i_partitioning          JSONB,
+    IN  i_id_checkpoint         UUID,
+    IN  i_checkpoint_name       TEXT,
+    IN  i_process_start_time    TIMESTAMP WITH TIME ZONE,
+    IN  i_process_end_time      TIMESTAMP WITH TIME ZONE,
+    IN  i_function_names        TEXT[],
+    IN  i_control_columns       TEXT[][],
+    IN  i_measure_values        JSONB[],
+    IN  i_by_user               TEXT,
```
lsulak, line 26: is this the checkpoint's author?

database/src/main/runs/checkpoints.ddl
```diff
@@ -13,21 +13,20 @@
  * limitations under the License.
  */
 
--- DROP TABLE IF EXISTS runs.checkpoints;
-
 CREATE TABLE runs.checkpoints
 (
     id_checkpoint           UUID NOT NULL,
-    key_segmentation        BIGINT NOT NULL,
+    fk_partitioning         BIGINT NOT NULL,
```
lsulak, line 19: don't forget to add author as well

database/src/main/runs/_get_id_partitioning.sql
benedeki: * yes the `_` at the beginning of the function name signs a "private" function.
* private function is not to be called outside, therefore not grants; thanks to that it can be lenient on input sanitation
* ownership automatically means execution right
* the magic of the function being (indirectly) usable by outside user is in the `SECURITY DEFINER` specification in the functions - it means that all the code within the function will be executed with right of the ROLE that owns the function not the ROLE that called the function. This prevents the needs to maintain cascade of access rights and opening _private_ functions to outside use.

database/src/main/runs/checkpoint_measure_definitions.ddl
benedeki: I wanted to have it the same as in DTO.

database/src/main/runs/checkpoints.ddl
benedeki: `created_by`
I know the name goes against what I wrote above. But this is widespread and rhymes with `created_at`, `updated_by` etc....

database/src/main/runs/checkpoints.ddl
```diff
@@ -13,21 +13,20 @@
...
-    workflow_name           TEXT,
     process_start_time      TIMESTAMP WITH TIME ZONE NOT NULL,
     process_end_time        TIMESTAMP WITH TIME ZONE,
     created_by              TEXT NOT NULL,
     created_at              TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
     CONSTRAINT segments_pk PRIMARY KEY (id_checkpoint)
 );
 
-CREATE INDEX checkpoints_idx1 ON runs.checkpoints (key_segmentation);
+COMMENT ON COLUMN runs.checkpoints.id_checkpoint is 'The UUID is coming from outside world to distinguish repeated entry from a re-run checkpoint';
```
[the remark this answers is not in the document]
benedeki: Ideally, all columns should have comments. But writing them like this is cumbersome. _Ultet_ should help

database/src/main/runs/future/write_measurement.sql
benedeki: Thanks I missed this one being unfinished 😊

database/src/main/runs/write_checkpoint.sql
benedeki: Yes. Might change it to `i_author` 🤔

database/src/main/runs/write_checkpoint.sql
benedeki, line 33:
```suggestion
-- Function: runs.write_checkpoint(9)
```

database/src/main/runs/future/write_measurement.sql
benedeki: Ah, it's in the future dir.

benedeki:
Still need to test, but generally should be ready.

database/src/main/postgres/runs/partitionings.ddl
```diff
@@ -13,18 +13,16 @@
...
     created_by              TEXT NOT NULL,
     created_at              TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
-    CONSTRAINT segmentations_pk PRIMARY KEY (id_segmentation)
+    CONSTRAINT segmentations_pk PRIMARY KEY (id_partitioning)
 );
 
-ALTER TABLE runs.segmentations
-    ADD CONSTRAINT segmentations_unq UNIQUE (segmentation);
+ALTER TABLE runs.partitionings
+    ADD CONSTRAINT segmentations_unq UNIQUE (partitioning);
```
salamonpavel, line 26: The constraint might also be renamed to partitioning_unq.

salamonpavel: We might want to consider indexing the partitioning JSONB column as it's being used in _get_id_partitioning function when looking up partitioning id.

benedeki: Unique constraint automatically means an index created.

database/src/main/postgres/flows/_create_flow.sql
```diff
@@ -0,0 +1,68 @@
...
+-- Status codes:
+--      11                  - Flow created
+--
+--
+-------------------------------------------------------------------------------
+DECLARE
+    _id_flow        BIGINT;
+    _flow_name      TEXT;
+BEGIN
+    --generating the id explicitly to use it it custom flow name if needed
```
lsulak, line 48:
```suggestion
    --generating the id explicitly to use, if custom flow name if needed
```

lsulak: or, I don't know, maybe this comment can be dropped, not sure if it's useful

database/src/main/postgres/flows/_add_to_parent_flows.sql
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2023 ABSA Group Limited
```
lsulak, line 2:
```suggestion
 * Copyright 2023 ABSA Group Limited
```

extremely tiny detail, but we said that the licence header should be consistant and should be of a year of repo creation, not current year. It's like this on 4 places in total

database/src/main/postgres/runs/partitionings.ddl
```diff
@@ -13,18 +13,16 @@
...
+CREATE TABLE runs.partitionings
 (
-    id_segmentation         BIGINT NOT NULL DEFAULT global_id(),
-    segmentation            HSTORE NOT NULL,
+    id_partitioning         BIGINT NOT NULL DEFAULT global_id(),
+    partitioning            JSONB NOT NULL, -- TODO add  partitioning validity check #69
     created_by              TEXT NOT NULL,
     created_at              TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
-    CONSTRAINT segmentations_pk PRIMARY KEY (id_segmentation)
+    CONSTRAINT segmentations_pk PRIMARY KEY (id_partitioning)
```
lsulak, line 22:
```suggestion
    CONSTRAINT partitioning_pk PRIMARY KEY (id_partitioning)
```

database/src/main/postgres/runs/_get_id_partitioning.sql
```diff
@@ -0,0 +1,55 @@
...
+--
+-- Function: runs._get_id_partitioning(3)
+--      Gets the id of the provided partitioning, if it exits
+--
+-- Parameters:
+--      i_partitioning      - partitioning to look for
+--      i_blocking          - flag signaling if to put a lock on the record
+--
+-- Returns:
+--      id_partitioning     - id of the segmentation if it exists, NULL otherwise
```
lsulak, line 32:
```suggestion
--      id_partitioning     - id of the partitioning if it exists, NULL otherwise
```

database/src/main/postgres/runs/write_checkpoint.sql
```diff
@@ -0,0 +1,112 @@
...
+    OUT status_text                 TEXT
+) RETURNS record AS
+$$
+-------------------------------------------------------------------------------
+--
+-- Function: runs.write_checkpoint(10)
+--      Creates a checkpoint and adds all the measurements that it consists of
+--
+-- Parameters:
+--      i_partitioning              - segmentation the measure belongs to
```
lsulak, line 36:
```suggestion
--      i_partitioning              - partitioning the measure belongs to
```

database/src/main/postgres/runs/measurements.ddl
```diff
@@ -13,20 +13,18 @@
...
 CREATE TABLE runs.measurements
 (
-    id_measurement                      UUID NOT NULL,
-    key_checkpoint_measure_definition   BIGINT NOT NULL,
-    key_checkpoint                      UUID NOT NULL,
-    value                               JSON NOT NULL,
+    id_measurement                      BIGINT NOT NULL DEFAULT global_id(),
+    fk_measure_definition               BIGINT NOT NULL,
+    fk_checkpoint                       UUID NOT NULL,
+    measurement_value                   JSONB NOT NULL,
```
lsulak, line 21: I think here we could even create a composite type, because we already know what the structure should be. What do you think? More flexibility can be good in some cases, but can bring unexpected situations/bad values in others.

Not now, I just wanted to understand your reasoning

database/src/main/postgres/runs/checkpoints.ddl
```diff
@@ -13,21 +13,21 @@
...
-    key_segmentation        BIGINT NOT NULL,
+    fk_partitioning         BIGINT NOT NULL,
     checkpoint_name         TEXT NOT NULL,
-    workflow_name           TEXT,
     process_start_time      TIMESTAMP WITH TIME ZONE NOT NULL,
     process_end_time        TIMESTAMP WITH TIME ZONE,
+    measured_by_atum_agent  BOOLEAN,
     created_by              TEXT NOT NULL,
     created_at              TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
     CONSTRAINT segments_pk PRIMARY KEY (id_checkpoint)
```
lsulak, line 26: probably not the best name for the constraint?

lsulak:
```suggestion
    CONSTRAINT checkpoint_pk PRIMARY KEY (id_checkpoint)
```
❓

database/src/main/postgres/runs/additional_data.ddl
```diff
@@ -13,12 +13,10 @@
...
 
--- DROP TABLE IF EXISTS runs.additional_data;
-
 CREATE TABLE runs.additional_data
 (
     id_additional_data      BIGINT NOT NULL DEFAULT global_id(),
-    key_segmentation        BIGINT NOT NULL,
+    fk_partitioning         BIGINT NOT NULL,
     ad_name                 TEXT NOT NULL,
     ad_value                TEXT,
```
lsulak, line 21: I thought that this will be a JSONB, why not?

database/src/main/postgres/runs/create_partitioning_if_not_exists.sql
```diff
@@ -0,0 +1,108 @@
...
+    _create_partitioning    BOOLEAN;
+    _status                 BIGINT;
+BEGIN
+
+    id_partitioning := runs._get_id_partitioning(i_partitioning, true);
+
+    _create_partitioning := id_partitioning IS NULL;
+
+    IF i_parent_partitioning IS NOT NULL THEN
+        -- TODO check if i_parent_partitioning is parent to i_partitioning #69
```
lsulak, line 58: I think that this comment can be dropped? We agreed thatpartitioning can be 'related' in child-parent way even if they don't share any 'keys', for example a join operation can cause this

database/src/main/postgres/runs/create_partitioning_if_not_exists.sql
```diff
@@ -0,0 +1,108 @@
...
+          -- copying measure definitions to establish continuity
+          INSERT INTO runs.measure_definitions(fk_partitioning, measure_name, measured_columns, created_by, created_at)
+          SELECT id_partitioning, CMD.measure_name, CMD.measured_columns, CMD.created_by, CMD.created_at
+          FROM runs.measure_definitions CMD
+          WHERE CMD.fk_partitioning = _fk_parent_partitioning;
+
+          -- additional data are not copied, they are specific for particular partitioning
+        ELSIF (_status = 11) THEN
+          status := 12;
+          status_text := 'Partitioning updated';
```
lsulak, line 98: I really don't know whether this is a good sentense - partitioning can't be updated, right? This is a situation when the parent partitioning is added to the flows

database/src/main/postgres/runs/_write_measurement.sql
```diff
@@ -0,0 +1,82 @@
...
+    INTO _fk_measure_definition;
+
+    IF NOT found THEN
+        INSERT INTO runs.measure_definitions (fk_partitioning, measure_name, measured_columns, created_by)
+        VALUES (i_fk_partitioning, i_measure_name, i_measured_columns, i_by_user)
+        RETURNING id_measure_definition
+        INTO _fk_measure_definition;
+
+        status := 11;
+        status_text := 'Measurement added including measurement definition';
```
lsulak, line 67:
```suggestion
        status_text := 'Measurement added including measure definition';
```

database/src/main/postgres/runs/_write_measurement.sql
```diff
@@ -0,0 +1,82 @@
...
+--      i_measurement_value     - value of the measure
+--      i_by_user               - user behind the change
+--
+-- Returns:
+--      status              - Status code
+--      status_text         - Status text
+--
+-- Status codes:
+--      10                  - OK
+--      11                  - Measurement added including measurement definition
```
lsulak, line 47:
```suggestion
--      11                  - Measurement added including measure definition
```

database/src/main/postgres/runs/_write_measurement.sql
```diff
@@ -0,0 +1,82 @@
...
+
+        status := 11;
+        status_text := 'Measurement added including measurement definition';
+    ELSE
+        -- assuming ok result
+        status := 10;
+        status_text := 'OK';
+    END IF;
+
+    INSERT INTO runs.measurements (fk_measure_definition, fk_checkpoint, measurement_value)
```
lsulak, line 74: what if the measurement already exist? I.e. this combination of measure definition and checkpoint. This can easily happen if this function is called twice. I think a constraint violation would be triggered

database/src/main/runs/_get_id_partitioning.sql
lsulak: awesome, thanks a lot for such detailed explanation!

database/src/main/postgres/runs/_write_measurement.sql
lsulak: Hmm okay then, I think it can be a good solution, we don't have to have all the functions idempotent / absolutely safe against such situations

database/src/main/postgres/runs/additional_data.ddl
lsulak: I'm not sure if I understood you to be honest, apologies. I was referring to `ad_name` and `ad_value` - so this means that a sinle record in `additional_data` table will hold only 1 metadata record that is 'flat' like this. I thought that a single additional_data record should be a JSON structure-check our data model:
```
case class AdditionalDataDTO(
  additionalData: Map[String, Option[String]]
)
```

lsulak:
File `list_flows.sql` from the `future/` dir still has `Copyright 2022 ABSA Group Limited`, it's the only one left, please if you can also change it so that the whole repo has it right

database/src/main/postgres/runs/additional_data.ddl
benedeki: It's decomposed for better handling. Indexing, searching, patterning and merging.
If the DB is to understand the data (do operations on them), I really prefer if they are not hidden in complex types.

lsulak: Okay then. Let's use 'flat' data types whenever possible. In case of partitioning and measurements - those could also be decomposed, but let's leave it all as is, since we'll need some flexibility given the current state of the project (and already existing implementation)

lsulak (approved):
- [x]  code reviewed
- [x]  pulled
- [x]  built
- [ ]  ran against PR - I'll do it once I'll finish #23, I don't want to test similar/same things multiple times at the moment

PR #86, opened 2023-10-12
Feature/54 checkpoint from server

### LINKED ISSUE #54: Read Checkpoint definition from the Server

salamonpavel:
Closes #54


agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -47,25 +47,41 @@ class AtumAgent private() {
    *  @return
    */
   def getOrCreateAtumContext(atumPartitions: AtumPartitions): AtumContext = {
-    contexts.getOrElse(atumPartitions, new AtumContext(atumPartitions, this))
+    getContextOrElse(
```
lsulak, line 50: I would maybe extract the `AtumContext.fromDTO()` call into a val so that it's not that nested

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -47,25 +47,41 @@ class AtumAgent private() {
...
+      )
+    )
   }
 
   def getOrCreateAtumSubContext(subPartitions: AtumPartitions)(implicit parentAtumContext: AtumContext): AtumContext = {
     val newPartitions: AtumPartitions = parentAtumContext.atumPartitions ++ subPartitions
-    getContextOrElse(newPartitions, parentAtumContext.copy(atumPartitions = newPartitions, agent = this))
+    getContextOrElse(
+      newPartitions,
+      AtumContext.fromDTO(
```
lsulak, line 63: potentially same comment as above, it can be extracted into val and not nested like this for code readability

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -47,25 +47,41 @@ class AtumAgent private() {
...
   }
 
-  private def getContextOrElse(atumPartitions: AtumPartitions, creationMethod: =>AtumContext): AtumContext = {
-    synchronized{
-      contexts.getOrElse(atumPartitions, {
-        val result = creationMethod
-        contexts = contexts + (atumPartitions -> result)
-        result
-      })
+  private def getContextOrElse(atumPartitions: AtumPartitions, creationMethod: => AtumContext): AtumContext = {
```
lsulak, line 73: I think that the `creationMethod` is not a good name - it's actually a context already. If I understood your code and intentions correctly, then you are first calling the method and then its result is being passed here - so this is actually something like `existingContext` ?

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
salamonpavel: Refactored.

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
salamonpavel: Refactored.

PR #88, opened 2023-10-16
scalafmt config more close to scala style guide

cerveada:


.scalafmt.conf
```diff
@@ -3,7 +3,7 @@ runner.dialect = scala212
 
 maxColumn = 120
 
-align.preset = some
+align.preset = none
```
lsulak, line 6: I quite liked `some` here, but I'd like to hear your opinion on this one (i.e. why `none` ? )

.scalafmt.conf
```diff
@@ -16,6 +16,5 @@ docstrings.style = AsteriskSpace
 docstrings.wrap = no
 docstrings.removeEmpty = true
 
-align.openParenDefnSite = true
-align.openParenCallSite = true
-
+align.openParenDefnSite = false
+align.openParenCallSite = false
```
lsulak, line 20: Nice, it works, I just tested it & I like it

.scalafmt.conf
cerveada: It definitely may look nice in some instances, but it creates unnecessary changes in diff. It also may create lines longer than the limit, where it would not happen without it.

Those are not huge issues, so I am ok using 'some' if you think it's worth it.

lsulak (approved):

PR #89, opened 2023-10-16
#87: DTOs for measures and measure results, MeasurementBuilder, Agent…

### LINKED ISSUE #87: Refactor Measure and MeasureResult class in model and agent
## Background
Currently the implementation of `MeasureResult` is only on the model side, and does not carry type information. Also, it's not used in Agent - we need to somehow connect those two modules.

## Feature
Improve the current `MeasureResult` class so that it carries information about type and also add support for having additional metadata / helper measures used - it can be an empty Map for now.

Also add a mapping between these DTO objects and internal types in the Agent - feel free to use `MeasuresMapper` for this, it might be what we need here.

## Proposed Solution 

```scala
case class MeasureResultDTO(
                           mainValue: TypedValue,
                           supportValues: Map[String, TypedValue] = Map.empty
                           )


object MeasureResultDTO {
  case class TypedValue(
                         value: String,
                         valueType: String //actually enumaration, but those can cause issues in in serde
                       )
}


//count -> MeasureResultDTO in JSON format
//"mainValue": {
//  "value": "103312",
//  "valueType": "Long"
//}
//"supportValues": {}


//avg -> MeasureResultDTO in JSON format
//"mainValue": {
//  "value": "2.5",
//  "valueType": "BigDecimal"
//}
//"supportValues": {
//   "count": {
//     "value": "2",
//      "valueType": "Long"
//    }
//   "total": {
//     "value": "5",
//      "valueType": "BigDecimal"
//    }
// }

.addLongMeasure("count", 103312)
```

salamonpavel:
Closes #87


agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -47,17 +47,20 @@ class AtumAgent private[agent] () {
...
 
   def getOrCreateAtumSubContext(subPartitions: AtumPartitions)(implicit parentAtumContext: AtumContext): AtumContext = {
     val newPartitions: AtumPartitions = parentAtumContext.atumPartitions ++ subPartitions
-    val atumContextDTO = dispatcher.getOrCreateAtumContext(
-      AtumPartitions.toPartitioning(newPartitions),
-      Some(AtumPartitions.toPartitioning(parentAtumContext.atumPartitions))
-    )
+
+    val newPartitionsDTO = AtumPartitions.toSeqPartitionDTO(newPartitions)
+    val maybeParentPartitionsDTO = Some(AtumPartitions.toSeqPartitionDTO(parentAtumContext.atumPartitions))
```
lsulak, line 60: actually this always will have the value, so perhaps I would rename it as

```suggestion
    val parentPartitionsDTO = Some(AtumPartitions.toSeqPartitionDTO(parentAtumContext.atumPartitions))
```

agent/src/main/scala/za/co/absa/atum/agent/model/MeasuresMapper.scala
```diff
@@ -17,24 +17,26 @@
 package za.co.absa.atum.agent.model
 
 import za.co.absa.atum.agent.model.Measure._
+import za.co.absa.atum.model.dto
 
 case class UnsupportedMeasureException(msg: String) extends Exception(msg)
```
lsulak, line 22: actually, consider (but I'll leave it up to you, just an idea) to move this to something like `za.co.absa.atum.agent.exception` module

agent/src/main/scala/za/co/absa/atum/agent/model/MeasuresMapper.scala
```diff
@@ -17,24 +17,26 @@
...
 object MeasuresMapper {
 
-  def mapToMeasures(measures: Set[za.co.absa.atum.model.Measure]): Set[za.co.absa.atum.agent.model.Measure] = {
+  def mapToMeasures(measures: Set[dto.MeasureDTO]): Set[za.co.absa.atum.agent.model.Measure] = {
     measures.map(createMeasure)
   }
 
-  private def createMeasure(measure: za.co.absa.atum.model.Measure): za.co.absa.atum.agent.model.Measure = {
+  private def createMeasure(measure: dto.MeasureDTO): za.co.absa.atum.agent.model.Measure = {
     val controlColumn = measure.controlColumns.head
```
lsulak, line 31: later we could extend this to handle situations where there are different number of control columns for different measures. But it's okay for now

agent/src/main/scala/za/co/absa/atum/agent/model/MeasurementBuilder.scala
```diff
@@ -0,0 +1,36 @@
+package za.co.absa.atum.agent.model
```
lsulak, line 1: licence is missing here

salamonpavel: Added missing licence info.

agent/src/main/scala/za/co/absa/atum/agent/model/MeasuresMapper.scala
salamonpavel: Moved

PR #91, opened 2023-10-18
#55: Checkpoint creation on provided data

### LINKED ISSUE #55: Create checkpoint based on provided data, no measuring
## Background
Sometimes control measure doesn't need to be measured, they are provided from outside.

## Feature
Add the ability to `AtumContext` and the implicit to just post the measurement values instead of measuring them. From that point they will behave like the measured ones though (calling same API etc.)

Depends on #26
Depends on #52

salamonpavel:
Closes #55

### DIFF
--- a/agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
+++ b/agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
@@ -44,7 +44,7 @@ object Measure {
   )
   val supportedMeasureNames: Seq[String] = supportedMeasures.map(_.measureName)
 
-  case class RecordCount private (controlCol: String, measureName: String, onlyForNumeric: Boolean) extends Measure { // extends MeasureType
+  case class RecordCount private (controlCol: String, measureName: String, onlyForNumeric: Boolean) extends Measure {
 
     override def function: MeasurementFunction =
       (ds: DataFrame) => ds.select(col(controlCol)).count().toString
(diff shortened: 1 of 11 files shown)


agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -40,44 +44,74 @@ object Measure {
   )
   val supportedMeasureNames: Seq[String] = supportedMeasures.map(_.measureName)
 
-  class MeasureType(val measureName: String, val onlyForNumeric: Boolean)
-
-  case object RecordCount extends MeasureType("count", false)
-  case class RecordCount(controlCol: String) extends Measure {
+  case class RecordCount private (controlCol: String, measureName: String, onlyForNumeric: Boolean) extends Measure { // extends MeasureType
```
lsulak, line 47:
```suggestion
  case class RecordCount private (controlCol: String, measureName: String, onlyForNumeric: Boolean) extends Measure {
```

agent/src/test/scala/za/co/absa/atum/agent/model/MeasurementBuilderTest.scala
```diff
@@ -0,0 +1,79 @@
...
+    val measurementDTO = MeasurementBuilder.buildMeasurementDTO(measurement)
+
+    val expectedTypedValue = TypedValue("stringValue", ResultValueType.String)
+
+    assert(measurementDTO.result.mainValue == expectedTypedValue)
+  }
+
+  "buildMeasurementDTO" should "throw exception for unsupported result value type" in {
+    val measure = SumOfValuesOfColumn("col")
+    val measurement = Measurement(measure, 1)
```
lsulak, line 74: I was wondering whether there should be some automatic conversion between int and possible derivable types...but it's probably better to be as strict and exact as possible, so let's leave it as is :)

lsulak (approved):
Good to go! I pulled, built, ran tests, reviewed, and consulted with Pavel a few times. Approved, nice job! :)

PR #95, opened 2023-10-25
(only idea!) #94 measure validation on column

### LINKED ISSUE #94: Agent: Validate Measure Applicability for a given DF (column)
## Background
As the Issue title says. For example, it doesn't make sense to calculate SUM on non-numeric column.

lsulak:
Validation of measure calculations on agent side.

Closes #94

Note: don't review this one please, it might even be disregarded. It was just an idea!


agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -34,18 +36,37 @@ class AtumContext private[agent] (
...
+) extends Logging {
 
   def currentMeasures: Set[Measure] = measures
 
   def subPartitionContext(subPartitions: AtumPartitions): AtumContext = {
     agent.getOrCreateAtumSubContext(atumPartitions ++ subPartitions)(this)
   }
 
+  private def validateMeasureApplicability(measure: Measure, df: DataFrame): Unit = {
+    require(
```
lsulak, line 48: todo change this to Try - here and below instead of the `log.warn`  as part of this ticket

lsulak: to be done later, I created a ticket for this

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -34,18 +36,37 @@ class AtumContext private[agent] (
...
-) {
+) extends Logging {
 
   def currentMeasures: Set[Measure] = measures
 
   def subPartitionContext(subPartitions: AtumPartitions): AtumContext = {
     agent.getOrCreateAtumSubContext(atumPartitions ++ subPartitions)(this)
   }
 
+  private def validateMeasureApplicability(measure: Measure, df: DataFrame): Unit = {
```
benedeki, line 47: Would make this a method of `Measure`, so it can actually be customized based on each function.

benedeki (commented):
Sorry, not sure if this have priority now 🤔

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
lsulak: Good idea actually. Measure can have this one implemented in it, and individual ones can override if needed. Especially useful for Measures that don't operate on any column, such as `count`

lsulak: changed

lsulak:
Probably not gonna do it like this

PR #96, opened 2023-10-25
#59: create adjust endpoint for reading checkpoint defintion of partioning

### LINKED ISSUE #59: Create (adjust) endpoint for reading checkpoint defintion of partioning
## Background
The REST API endpoints of the Server

## Feature
Create (or adjust if already exists) API endpoint that will return the checkpoint definition (its defined measures)

Depends on #37

TebaleloS:
Retrieve checkpoint endpoint
* Defined endpoint for creating partitioning.
* Prepared the service method that will be used to create parting if it doesn't exist.
* Added a case class that defines the fields that will be used to retrieve the checkpoint definition when creating the partitioning.

Closes #59


server/src/main/scala/za/co/absa/atum/server/api/service/DatabaseService.scala
```diff
@@ -38,4 +37,13 @@ class DatabaseService @Autowired()() {
     } yield checkpoint
   }
 
+  /**
+   * Function to retrieve checkpoint based on the provided fields
+   *
+   * @param filterCriteria JSON object containing the fields for filtering the checkpoint
```
lsulak, line 43: what is this filterCriteria about? I saw it elsewhere as well

server/src/main/scala/za/co/absa/atum/server/api/service/DatabaseService.scala
```diff
@@ -38,4 +37,13 @@ class DatabaseService @Autowired()() {
...
+   */
+  def createPartitioningIfNotExists(partitioningInfo: PartitioningDTO): Seq[PartitionDTO] = {
```
lsulak, line 45: I think that the function you are trying to call here is not gonna return PartitionDTO sequence, but something else, check the DB function definition:

```
CREATE OR REPLACE FUNCTION runs.create_partitioning_if_not_exists(
    IN  i_partitioning          JSONB,
    IN  i_by_user               TEXT,
    IN  i_parent_partitioning   JSONB = NULL,
    OUT status                  INTEGER,
    OUT status_text             TEXT,
    OUT id_partitioning         BIGINT
)
```

server/src/test/scala/za/co/absa/atum/server/api/service/DatabaseServiceTest.scala
```diff
@@ -0,0 +1,28 @@
...
+
+    val results = databaseService.saveCheckpoint(checkpointDTO)
+    println(results)
+    results.complete(checkpointDTO)
+
+  }
+
+  test("testReadCheckpoint") {}
+
+  test("testPostgresAccessProvider") {}
```
lsulak, line 26: this doesn't really test anything, is it still in progress?

lsulak:
I really need this so I fixed the merge conflicts and did some refactoring. Apologies @TebaleloS, I know that it's not a good practice to do this, but I want to work on the #23 during the weekend a bit.

I also removed the Partitioning Created DTO you created, because PartitioningDTO is perfectly suitable for the job now I think. If something is missing, we can add it there, not creating a new DTO that is almost the same as the old one.

lsulak:
The build is failing because of the unit tests in `DatabaseServiceTest` - when I commented it out, it was fully working

server/src/test/scala/za/co/absa/atum/server/api/service/DatabaseServiceTest.scala
TebaleloS: Yes, it was just a preparation and I forgot to remove them because the implementation was totally wrong.

server/src/main/scala/za/co/absa/atum/server/api/service/DatabaseService.scala
TebaleloS: It has being changed to `partitioningInfo`.

lsulak:
Closing, replaced by https://github.com/AbsaOSS/atum-service/pull/121 and future PR related to the DB tests

PR #104, opened 2023-11-01
#105: Refactoring Measurement, from Agent side

### LINKED ISSUE #105: Agent: Refactor Measurements
## Background
Currently the Measurements and MeasurementResults are a bit unclean and not unified. Try to find a way how to refactor it - also MeasurementProcessor has some class related to it that could be put elsewhere.

lsulak:
Addressing `Measurement` and `MeasurementResult` - I tried to make it much more simpler and added some validations even. Now `Measure` contains only 2 attributes; the result itself is separated.

I improved the MeasurementBuilder and now the `Measures` and `Measurements` are `Set` but with some additional validation on uniqueness - i.e. a user can't define the same Measures against the same columns with different result values - it's not a valid case.

Closes #105


TebaleloS (commented):
This looks good. The only problem is that it cannot compile successfully, it has a type mismatch on `found   : scala.collection.immutable.Set[String]
[error]  required: Seq[Any]
[error]       val measureNames = scalaIterableToSQLArray(values.measurements.map(_.measure.measureName))`

lsulak:
> This looks good. The only problem is that it cannot compile successfully, it has a type mismatch on `found : scala.collection.immutable.Set[String] [error] required: Seq[Any] [error] val measureNames = scalaIterableToSQLArray(values.measurements.map(_.measure.measureName))`

Fixed, thanks, it's already in master now

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -142,7 +141,7 @@ class AtumContext private[agent] (
   /**
    * Adds a measure to the AtumContext.
    *
-   * @param measure the measure to be added
+   * @param newMeasure the measure to be added
```
salamonpavel, line 144: `measure` is better then `newMeasure`, it's obvious from the context that it's a new measure. I suggest to rename the parameter to `measure`.

salamonpavel: I wonder if we could use vals instead of vars in AtumContext ...

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
```diff
@@ -61,13 +59,13 @@ class HttpDispatcher(config: Config) extends Dispatcher with Logging {
 
     val response = backend.send(request)
 
-    safeResponseBody(response).get
+    safeResponseBody(response)
   }
 
-  private def safeResponseBody(response: Response[Either[String, String]]): Try[String] = {
+  private def safeResponseBody(response: Response[Either[String, String]]): String = {
```
salamonpavel, line 65: Absolutely, I like the change, going from Either to Try and then calling .get is an anti-pattern. Throwing exceptions can sometimes be disruptive and it's often better to let the client code decide how to handle error cases. I wonder if we could instead return Either[HttpException, String]. This way we would be using type system to explicitly signal that the operation might fail with an HttpException. We could also have two flavours of saveCheckpoint method (one throwing the exception, another returning Either). This would give the client code flexibility to choose the version that best fits their error handling strategy.

salamonpavel: Btw I am not sure about the method's name. Can you think of sth better?

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -48,86 +47,86 @@ object AtumMeasure {
...
 
-  case class SumOfValuesOfColumn private (measureName: String, controlCol: String) extends AtumMeasure {
+  case class SumOfValuesOfColumn private (measureName: String, measuredCol: String) extends AtumMeasure {
     private val columnAggFn: Column => Column = column => sum(column)
 
-    override def function: MeasurementFunction = (ds: DataFrame) => {
-      val dataType = ds.select(controlCol).schema.fields(0).dataType
-      val resultValue = ds.select(columnAggFn(castForAggregation(dataType, col(controlCol)))).collect()
-      ResultOfMeasurement(handleAggregationResult(dataType, resultValue(0)(0)), resultValueType)
+    override def function: MeasurementProcessor.MeasurementFunction = (ds: DataFrame) => {
```
salamonpavel, line 86: Why do you prefer `MeasurementProcessor.MeasurementFunction` over `MeasurementFunction`?

agent/src/main/scala/za/co/absa/atum/agent/model/MeasureResult.scala
```diff
@@ -0,0 +1,87 @@
...
+package za.co.absa.atum.agent.model
+
+import za.co.absa.atum.agent.exception.AtumAgentException.MeasurementException
+import za.co.absa.atum.model.dto.MeasureResultDTO.ResultValueType
+
+/**
+ *  This trait defines a contract for a measure result.
+ */
+trait MeasureResult {
+  val resultValue: Any
```
salamonpavel, line 26: Isn't better to use generic `T` instead of `Any`? If we were to use generic `T` we would preserve type safety (`Any` can be any type at all) and flexibility (you cannot apply constraints like contexts bounds to `Any`).

agent/src/main/scala/za/co/absa/atum/agent/model/MeasurementBuilder.scala
```diff
@@ -16,19 +16,40 @@
...
-  private [agent] def buildMeasurementDTO(measurement: Measurement): MeasurementDTO = {
-    val measureName = measurement.measure.measureName
-    val measureDTO = MeasureDTO(measureName, measurement.measure.controlColumns)
-    val measureResultDTO = MeasureResultDTO(TypedValue(measurement.resultValue.toString, measurement.resultType))
+  private[agent] def validateMeasureUniqueness(measurements: Set[Measurement]): Unit = {
+    val allMeasures = measurements.toSeq.map(_.measure)
 
+    val originalMeasuresCnt = allMeasures.size
+    val uniqueMeasureColumnCombinationCnt = allMeasures.map(m =>
+      Tuple2(m.measureName, m.measuredColumns)  // there can't be two same measures defined on the same column(s)
```
salamonpavel, line 32: The `Tuple2` is not needed.

agent/src/main/scala/za/co/absa/atum/agent/model/MeasurementBuilder.scala
```diff
@@ -16,19 +16,40 @@
...
 /**
  * This object provides a functionality to convert a measurement to its DTO representation.
  */
 private [agent] object MeasurementBuilder {
 
-  private [agent] def buildMeasurementDTO(measurement: Measurement): MeasurementDTO = {
-    val measureName = measurement.measure.measureName
-    val measureDTO = MeasureDTO(measureName, measurement.measure.controlColumns)
-    val measureResultDTO = MeasureResultDTO(TypedValue(measurement.resultValue.toString, measurement.resultType))
+  private[agent] def validateMeasureUniqueness(measurements: Set[Measurement]): Unit = {
```
salamonpavel, line 27: Do you need to really validate this? Isn't that what `Set` itself does as it uses equals and hashCode methods.

agent/src/main/scala/za/co/absa/atum/agent/model/MeasurementBuilder.scala
salamonpavel: Maybe it could be simplified ...

```scala
if (measurements.size != measurements.toSeq.distinctBy(m => (m.measure.measureName, m.measure.measuredColumns)).size) {
      throw MeasurementException(
        s"Measure and measuredColumn combinations must be unique, i.e. they cannot repeat!"
      )
    }
```

salamonpavel: Did you think of any other approaches? `Set` uses `equals` and `hashCode` methods to determine uniqueness. You could overwrite those on `AtumMeasure` and no validation method would be needed.

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
lsulak: No good idea and I can't remember, I would say that it's unintentional side effect of refactoring. I'll certainly change it

agent/src/main/scala/za/co/absa/atum/agent/model/MeasureResult.scala
lsulak: Yes I'm absolutely aware of that. The thing is that if this would be of a generic type, then this trait would need to be of a generic type as well, and every usages of it would need to take that into an account. I gave it a try (1-2 hours) twice in the past and I didn't like the solution at all. I even wanted to include `Shapeless` here and this would be a the first step, but as I said I decided to keep it like this for now

agent/src/main/scala/za/co/absa/atum/agent/model/MeasurementBuilder.scala
lsulak: Hmm I was thinking about moving this code more to `AtumMeasure`, but in the end, the Set that the user specifies is not a set of AtumMeasures, but a Set of `Measurement` objects. So I thought that the validation method is warranted - and it's relatively 'fail-early' approach, since this would be executed during CheckpointDTO creation. But I still like your idea about using `distinctBy` method :)

lsulak: On the other hand it might make sense, technically speaking, to implement something like `equals` and `hashCode` directly on `Measurement`, but I feel like a direct, dedicated validation, that's extracted elsewhere, thus not changing the 'default' behavior of case class uniqueness in this case, seems like a better idea

lsulak: Hmm actually `distinctBy` is not generally available in Scala 2.12 IIRC - I'll still think about some improvements for sure

salamonpavel: Interesting, I had no idea it's 2.13 addition. I agree with your point about not changing the default behaviour.

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
lsulak: changed

agent/src/main/scala/za/co/absa/atum/agent/model/MeasureResult.scala
salamonpavel: @TebaleloS @benedeki 
I am interested in opinion of others on this matter.

agent/src/main/scala/za/co/absa/atum/agent/model/MeasureResult.scala
```diff
@@ -0,0 +1,87 @@
...
+   * These adjustments are needed to be performed - to avoid some floating point issues
+   * (overflows, consistent representation of numbers - whether they are coming from Java or Scala world, and more).
+   */
+  case class MeasureResultByAtum(resultValue: String, resultValueType: ResultValueType.ResultValueType) extends MeasureResult
+
+  /**
+   * When the application/user provides the actual results by himself, the type is precise and we don't need
+   * to do any adjustments.
+   */
+  case class MeasureResultProvided[T](resultValue: T, resultValueType: ResultValueType.ResultValueType) extends MeasureResult
```
benedeki, line 49: Why both the classes, when the former is basically a subset of later?

benedeki, line 49: If kept would make the constructor private to protect against inconsistent `T` and `ResultValueType.ResultValueType`

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -16,81 +16,26 @@
...
-  object MeasurementProvided {
-
-    private def handleSpecificType[T](
-      measure: AtumMeasure,
-      resultValue: T,
-      requiredType: ResultValueType.ResultValueType
-    ): MeasurementProvided[T] = {
+  private def validateMeasurement(measure: AtumMeasure, result: MeasureResult): Unit = {
+    val actualType = result.resultValueType
+    val requiredType = measure.resultValueType
```
benedeki, line 33: This lead to runtime error. Would prefer to ensure it cannot happen. Isn't that possible?

agent/src/main/scala/za/co/absa/atum/agent/model/MeasurementBuilder.scala
```diff
@@ -16,19 +16,40 @@
...
+    ).distinct.size
+
+    if (originalMeasuresCnt != uniqueMeasureColumnCombinationCnt)
+      throw MeasurementException(
+        s"Measure and measuredColumn combinations must be unique, i.e. they cannot repeat! Got: $allMeasures"
+      )
+  }
+
+  private[agent] def buildMeasurementsDTO(measurements: Set[Measurement]): Set[MeasurementDTO] = {
+    validateMeasureUniqueness(measurements)
```
benedeki, line 42: Why the validation? If not set, duplicates will be eliminated. So what?

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
```diff
@@ -61,13 +59,13 @@ class HttpDispatcher(config: Config) extends Dispatcher with Logging {
...
-    safeResponseBody(response).get
+    handleResponseBody(response)
   }
 
-  private def safeResponseBody(response: Response[Either[String, String]]): Try[String] = {
+  private def handleResponseBody(response: Response[Either[String, String]]): String = {
     response.body match {
-      case Left(body) => Failure(HttpException(response.code.code, body))
-      case Right(body) => Success(body)
+      case Left(body) => throw HttpException(response.code.code, body)
```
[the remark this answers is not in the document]
salamonpavel: It's unnecessary conversion. If you already have an Either, there's no need to convert it to a Try. There is no benefit in doing that.

agent/src/main/scala/za/co/absa/atum/agent/model/MeasureResult.scala
salamonpavel: @lsulak 
Can you maybe elaborate a bit more on the problems the generic would bring? I understand that the type parameter will be reflected in signatures throughout the code but when you create an instance of a class you don't need to specify the type as it's inferred.

agent/src/main/scala/za/co/absa/atum/agent/model/MeasurementBuilder.scala
lsulak: Depends what you mean by duplicates. Some scenarios will be eliminated, some not, I want to cover it all.

These duplicates would be eliminated by the nature of set being... set :)

```scala
      Measurement(RecordCount(), MeasureResult("1", ResultValueType.Long)),
      Measurement(SumOfValuesOfColumn("col1"), MeasureResult(BigDecimal(1.2))),
      Measurement(SumOfValuesOfColumn("col1"), MeasureResult(BigDecimal(1.2))),
      Measurement(DistinctRecordCount(Seq("col2")), MeasureResult(3L))
```
^ this is actually 3 unique records

--

but, this is actually 4 records:

```
      Measurement(RecordCount(), MeasureResult("1", ResultValueType.Long)),
      Measurement(SumOfValuesOfColumn("col1"), MeasureResult(BigDecimal(1.2))),
      Measurement(SumOfValuesOfColumn("col1"), MeasureResult(BigDecimal(1.4))),
      Measurement(DistinctRecordCount(Seq("col2")), MeasureResult(3L))
```

so there is a 'conflict' that the SET can't fix - second and third measurements are supposed to be the same - same Atum Context captured the same Measure on the same column but with different results. So I need to 'dig deeper' into the Set of Measurement objects - I need to check all Measures and their columns, and I need to make sure that only unique combination of these is further used

after all, there are unit tests testing (validity and non-valid cases) all this

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
lsulak: It absolutely is possible I believe - but the changes we would need to implement all begin with the measurement result value being `T` instead of `Any` - https://github.com/AbsaOSS/atum-service/pull/104#discussion_r1467322707, and I decided not to do it here.

The second step can be to use something like Shapeless and implement a combination of all measures and their data types that they support (I mean, just an implicit functions, based on example Filip provided)

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -51,83 +51,83 @@ object AtumMeasure {
...
     override val resultValueType: ResultValueType.ResultValueType = ResultValueType.Long
   }
   object RecordCount {
     private[agent] val measureName: String = "count"
     def apply(): RecordCount = RecordCount(measureName)
   }
 
-  case class DistinctRecordCount private (measureName: String, controlCols: Seq[String]) extends AtumMeasure {
-    require(controlCols.nonEmpty, "At least one control column has to be defined.")
+  case class DistinctRecordCount private (measureName: String, measuredCols: Seq[String]) extends AtumMeasure {
```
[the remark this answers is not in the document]
lsulak: Sure, I have no problems with that

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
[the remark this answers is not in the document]
lsulak: Exactly, I see it as one alternative to err handling being converted to another one, just didn't seemed valuable

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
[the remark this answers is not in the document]
lsulak: oh actually, there is this constructor attribute and then there is a def `override def measuredColumns`, so there would be a collision if I rename everything to `measuredColumns`

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -64,7 +64,7 @@ class AtumAgent private[agent] () {
   }
 
   /**
-   *  Provides an AtumContext given a `AtumPartitions` instance. Retrieves the data from AtumService API.
```
lsulak, line 67: https://docs.scala-lang.org/style/scaladoc.html

agent/src/main/scala/za/co/absa/atum/agent/model/MeasureResult.scala
lsulak:
```scala
trait MeasureResult[T] {
  val resultValue: T
  val resultValueType: ResultValueType.ResultValueType
}
```

basically this ^ would mean, that I would need to adjust code in `MeasurementProcessor` and its usecases, `Measurement` and its usecases, `MeasurementBuilder` and its usecases, including things like `AtumContext` -> `takeMeasurements` and then also unit tests and god knows what else :D 

I really like David's solution though, the whole 'generic' problem is nicely 'encapsulated' within the whole module

salamonpavel: I think you are actually right about this one. Anyway I have one suggestion, to replace `Enumeration` with `ADT` as shown below. Using sealed traits and case classes for `ResultValueType` is more idiomatic Scala and provides more flexibility. In this setup, `ResultValueType` is a sealed trait, which means it can only be extended in the same file. `StringResult`, `LongResult`, `BigDecimalResult` and `DoubleResult` are case classes the extend `ResultValueType`. This limits the possible types of `ResultValueType` to these four case classes.

This approach provides more flexibility because you can add methods and fields to the case classes if needed. It also provides more type safety because the compiler can check that you've handled all possible cases when you use pattern matching.
 
```scala
sealed trait ResultValueType
case class StringResult(value: String) extends ResultValueType
case class LongResult(value: Long) extends ResultValueType
case class BigDecimalResult(value: BigDecimal) extends ResultValueType
case class DoubleResult(value: Double) extends ResultValueType

trait MeasureResult {
  val resultValue: ResultValueType
}
```

lsulak: I don't know. I know that this is a common Scala pattern, but the original solution with the Enumerate is extremely small, really easy to refactor and extend if needed:

```scala
  object ResultValueType extends Enumeration {
    type ResultValueType = Value
    val String, Long, BigDecimal, Double = Value
```

salamonpavel: I leave it up to you to consider these options and decide. My personal preference would be the idiomatic solution with sealed trait and case classes. The reasoning was mentioned above, plus you wouldn't have to have a separate field just to keep track of the type.

salamonpavel (approved):

benedeki (dismissed):
Let's merge it. A major improvement.

.github/workflows/build.yml
```diff
@@ -23,23 +23,54 @@ on:
...
         with:
-          java-version: "adopt@1.8"
+          java-version: "adopt@1.11.0-11"
+
+      - name: Prepare testing database
+        run: sbt flywayMigrate
+
       - name: Build and run tests
-        run: sbt ++${{matrix.scala}} test doc
+        run: sbt "project server" ++${{matrix.scala}} test doc
```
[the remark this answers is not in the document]
lsulak: https://github.com/AbsaOSS/atum-service/issues/155

benedeki (approved):
Let's get it in 😉

PR #108, opened 2023-11-01
Feature/99 custom measure 2

### LINKED ISSUE #99: Add a possibility to let the user specify a custom Measure
## Background
This will be slightly different to the ones we have now. For this one, there would be no function body, just for specifying a custom Measure with provided data, and functionName (or measureName) that should 'label' the value properly.

salamonpavel:
Closes #99 
Added functionality to create custom measure.


agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -37,8 +37,8 @@ object Measurement {
   object MeasurementProvided {
 
     private def handleSpecificType[T](
-                               measure: Measure, resultValue: T, requiredType: ResultValueType.ResultValueType
-                             ): MeasurementProvided[T] = {
+                                       measure: AtumMeasure, resultValue: T, requiredType: ResultValueType.ResultValueType
+                                     ): MeasurementProvided[T] = {
```
cerveada, line 42: To keep the formatting consistent:
```scala
private def handleSpecificType[T](
  measure: AtumMeasure, 
  resultValue: T, 
  requiredType: ResultValueType.ResultValueType
): MeasurementProvided[T] = {
```
Generally it should be either all in one line or like this

lsulak: I agree with Adam

salamonpavel: Code has been reformatted.

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -68,13 +70,32 @@ object Measurement {
...
-  }
 
-  /**
-   * When the Atum Agent itself performs the measurements, using Spark, then in some cases some adjustments are
-   * needed - thus we are converting the results to strings always - but we need to keep the information about
-   * the actual type as well.
-   */
-  case class MeasurementByAtum(measure: Measure, resultValue: String, resultType: ResultValueType.ResultValueType)
-    extends Measurement
+    def forCustomMeasure[T](measureName: String, controlCol: String, resultValue: T): MeasurementProvided[T] = {
```
lsulak, line 74: I'm probably missing something but I think that this function is not needed at all - the only difference (apart from the definition, but that can be easily changed) between this one and `apply` is that internal `handleSpecificType` - I would perhaps try to refactor this one instead. Maybe drop the whole `handleSpeficType` function altogether and/or move it elsewhere, check how I did it here: https://github.com/AbsaOSS/atum-service/pull/104/files#diff-c65609cc426a47499c22c7fde4375cb3e305ed9e0e601c8c6b8dfd2c72b38899

lsulak: But those two functions are almost the same let's not violate DRY principle

lsulak (dismissed):
LGTM, as I said above, I think `Measurements.scala` can be much more improved, but I'm happy to do it here: https://github.com/AbsaOSS/atum-service/pull/104 so that you can merge this one. I'll leave it up to you :)

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
salamonpavel: Well, the handleSpecificType method on MeasurementProvided object expects to have resultValueType which is not really defined for CustomMeasure. If we change the CustomMeasure so that the client code will define it then we could use the mentioned method. But I don't quite see why would we do that cause in my opinion that would just complicate the creation of CustomMeasure. Moreover the client code cares about creation of MeasurementProvided, not the custom measure itself. For that reason I construct the MeasurementProvided in the forCustomMeasure method and use the provided result value's to determine resultValueType (for the MeasurementProvided). From the client code perspective I feel that's just a simplest way how to create a MeasurementProvided. 

```scala
MeasurementProvided.forCustomMeasure("customMeasureName", "col", BigDecimal(1))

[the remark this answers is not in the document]
Comment (salamonpavel) [reply agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala:74]: Please see the comment above.

**Comment by salamonpavel** [comment]
> LGTM, as I said above, I think `Measurements.scala` can be much more improved, but I'm happy to do it here: #104 so that you can merge this one. I'll leave it up to you :)

I feel like we keep refactoring the same things over and over again. (And more to come - multi column support etc.) :)

File: agent/src/main/scala/za/co/absa/atum/agent/model/MeasuresBuilder.scala [hunk 3]
```diff
@@ -17,16 +17,16 @@
 package za.co.absa.atum.agent.model
 
 import za.co.absa.atum.agent.exception.MeasureException
-import za.co.absa.atum.agent.model.Measure._
+import za.co.absa.atum.agent.model.AtumMeasure._
 import za.co.absa.atum.model.dto
```
TebaleloS, line 21: Why not importing `MeasureDTO` straight instead of importing the whole package while using only module.

TebaleloS (commented):
[1] Tested
[1] Compiled
[1] Reviewed and other than my comment everything looks good.

* I will approve after that. Good work!!!

*

agent/src/main/scala/za/co/absa/atum/agent/model/MeasuresBuilder.scala
salamonpavel: Thanks, I didn't notice it. Changed the import as per your suggestion.

lsulak:
> > LGTM, as I said above, I think `Measurements.scala` can be much more improved, but I'm happy to do it here: #104 so that you can merge this one. I'll leave it up to you :)
> 
> I feel like we keep refactoring the same things over and over again. (And more to come - multi column support etc.) :)

I agree, because some things are just changing and thus 'older' data model is not perfectly suitable anymore :D

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
lsulak: I agree but I still think there is a way. But why not to at least change

`def forCustomMeasure[T](measureName: String, controlCol: String, resultValue: T): MeasurementProvided[T]`

to

`def apply[T](measure: CustomMeasure, resultValue: T): MeasurementProvided[T]`

❓ 

It would be closer to the `def apply[T](measure: AtumMeasure, resultValue: T): Measurement` 'API'

salamonpavel: Renamed the forCustomMeasure method to apply.

agent/src/test/scala/za/co/absa/atum/agent/AtumContextTest.scala
```diff
@@ -108,7 +108,7 @@ class AtumContextTest extends AnyFlatSpec with Matchers {
     val measurements = Seq(
       MeasurementProvided(RecordCount("col"), 1L),
       MeasurementProvided(SumOfValuesOfColumn("col"), BigDecimal(1)),
-      MeasurementProvided.forCustomMeasure("customMeasureName", "col", BigDecimal(1))
+      MeasurementProvided("customMeasureName", "col", BigDecimal(1))
```
lsulak, line 111: I'm thinking that maybe the constructor of `CustomMeasure` don't have to be private, I think this seems nicer:

MeasurementProvided(CustomMeasure("customMeasureName", "col"), BigDecimal(1))

salamonpavel: Ok. Changed.

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -27,115 +27,90 @@ import za.co.absa.spark.commons.implicits.StructTypeImplicits.StructTypeEnhancem
...
         ResultOfMeasurement(resultValue, resultValueType)
       }
-  }
-  object RecordCount extends MeasureType {
-    def apply(controlCol: String): RecordCount = RecordCount(controlCol, measureName, resultValueType)
 
-    override val measureName: String = "count"
     override val resultValueType: ResultValueType.ResultValueType = ResultValueType.Long
   }
+  object RecordCount {
```
miroslavpojer, line 63: Missing empty line.

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -27,115 +27,90 @@ import za.co.absa.spark.commons.implicits.StructTypeImplicits.StructTypeEnhancem
...
-  }
-  object DistinctRecordCount extends MeasureType {
-    def apply(controlCol: String): DistinctRecordCount = {
-      DistinctRecordCount(controlCol, measureName, resultValueType)
-    }
 
-    override val measureName: String = "distinctCount"
     override val resultValueType: ResultValueType.ResultValueType = ResultValueType.Long
   }
+  object DistinctRecordCount {
```
miroslavpojer, line 78: Same, no empty line. Found multiple time below.

miroslavpojer (commented):
- [ ] pull
- [ ] build
- [ ] code review
- [ ] run tests
- [ ] check code coverage

lsulak (commented):
There are some merge conflicts but otherwise LGTM

miroslavpojer:
I have removed outdated JaCoCo code coverage post as they were for another version of Scala - change in title.
So removed to avoid confusion.

miroslavpojer (approved):

TebaleloS (commented):
Everything looks awesome!!! Just one suggestion, that you add at least add Scala doc to the methods for future reference, and also please resolve the conflicts.

PR #111, opened 2023-11-03
Feature/100 0 n columns

### LINKED ISSUE #100: Agent: Refactor the code, Measures requiring 0-n columns, not exactly 1
## Background
Check the internal Agent's code, especially the `Measure` trait - it has `controlCol` which is a String - but the MeasureDTO has a Sequence of these columns. DTO is correct - some measures can be using 0 columns and some can be using n columns - the goal of this ticket is to refactor the Agent's code and its internal model if needed; DTO probably won't need t be changed.

salamonpavel:
Closes #100 
Introduces support for 0-n columns (depends on individual measure).


agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -17,182 +17,156 @@
...
-  private val valueColumnName: String = "value"
+object AtumMeasure {
 
-  val supportedMeasures: Seq[MeasureType] = Seq(
-    RecordCount,
-    DistinctRecordCount,
-    SumOfValuesOfColumn,
-    AbsSumOfValuesOfColumn,
-    SumOfHashesOfColumn
+  val supportedMeasureNames: Seq[String] = Seq(
```
lsulak, line 44: why not to keep that? So we would offer strings as well as objects. No criticism, just want to know your reasoning

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -17,182 +17,156 @@
...
-import za.co.absa.spark.commons.implicits.StructTypeImplicits.StructTypeEnhancements
 
 /**
  *  Type of different measures to be applied to the columns.
  */
-sealed trait Measure extends MeasurementProcessor with MeasureType {
-  val controlCol: String
+sealed trait Measure {
+  val measureName: String
+  def controlColumns: Seq[String]
```
lsulak, line 31: I would even dare to say, let's rename it to `measuredColumns`. @benedeki WDYT?

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -17,182 +17,156 @@
...
+
+  case class DistinctRecordCount private (measureName: String, controlCols: Seq[String]) extends AtumMeasure {
+    require(controlCols.nonEmpty, "At least one control column has to be defined.")
 
-  case class DistinctRecordCount private (
-    controlCol: String,
-    measureName: String,
-    resultValueType: ResultValueType.ResultValueType
-  ) extends Measure {
+    private val columnExpression = countDistinct(col(controlCols.head), controlCols.tail.map(col): _*)
```
lsulak, line 72: I'm still not sure if this is what we want. I thought that it's about 'distinct -> count' operation, performed on a single column. Do you remember what was used in the old Atum & what can our users require @benedeki ?

lsulak:
> Code is based off of feature/99-custom-measure-2.

Why not to repoint the PR then? It would be easier to do the review

salamonpavel:
> Why not to repoint the PR then? It would be easier to do the review

Good point, will change the target branch to be the feature/99-custom-measure-2.

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
salamonpavel: With the suggested code it can be performed on single column or multiple columns. If this particular measure is expected to work differently please let me know. @benedeki

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
salamonpavel: I don't mind having the previous implementation once we really use that seq of objects. But until then it complicates matters (compilation) when one is refactoring the code.

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
lsulak: Ok we agreed that it's what we want, but not now and here, but in https://github.com/AbsaOSS/atum-service/issues/116

agent/src/main/scala/za/co/absa/atum/agent/model/Measurement.scala
```diff
@@ -74,13 +74,13 @@ object Measurement {
...
+          MeasurementProvided[T](CustomMeasure(measureName, Seq(controlCol)), resultValue, ResultValueType.Long)
         case _: Double =>
-          MeasurementProvided[T](CustomMeasure(measureName, controlCol), resultValue, ResultValueType.Double)
+          MeasurementProvided[T](CustomMeasure(measureName, Seq(controlCol)), resultValue, ResultValueType.Double)
         case _: BigDecimal =>
-          MeasurementProvided[T](CustomMeasure(measureName, controlCol), resultValue, ResultValueType.BigDecimal)
+          MeasurementProvided[T](CustomMeasure(measureName, Seq(controlCol)), resultValue, ResultValueType.BigDecimal)
         case _: String =>
-          MeasurementProvided[T](CustomMeasure(measureName, controlCol), resultValue, ResultValueType.String)
+          MeasurementProvided[T](CustomMeasure(measureName, Seq(controlCol)), resultValue,  ResultValueType.String)
```
lsulak, line 83:
```suggestion
          MeasurementProvided[T](CustomMeasure(measureName, Seq(controlCol)), resultValue, ResultValueType.String)
```

lsulak (commented):
I finished the review, I like the changes. There is only one tiny minor thing I found, and 1 thing regarding distinct count functionality. Otherwise LGTM and I will be happy to approve once these are addressed.

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
lsulak: Okay

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
benedeki: I like the universality of supporting multi-column uniqueness 👍

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -50,122 +49,124 @@ object AtumMeasure {
...
-  object AbsSumOfValuesOfColumn  {
-    private [agent] val measureName: String = "absAggregatedTotal"
+  object AbsSumOfValuesOfColumn {
+    private[agent] val measureName: String = "absAggregatedTotal"
     def apply(controlCol: String): AbsSumOfValuesOfColumn = AbsSumOfValuesOfColumn(measureName, controlCol)
   }
 
   case class SumOfHashesOfColumn private (measureName: String, controlCol: String) extends AtumMeasure {
-
+    private val columnExpression: Column = sum(crc32(col(controlCol).cast("String")))
```
benedeki, line 123: Could this function also support a list of columns, concatenated.
Extra bonus is `*` for all columns. 😉

benedeki (commented):
**To consider:**
I like class hierarchies. What you think of having 3 abstract classes
* `MeasureWithoutColumn`
* `MeasureOneColumn`
* `MeasureMultipleColumns`

agent/src/test/scala/za/co/absa/atum/agent/model/MeasurementTest.scala
```diff
@@ -28,15 +28,15 @@ class MeasurementTest extends AnyFlatSpec with Matchers with SparkTestBase { sel
 
   "MeasurementProvided" should "be able to be converted to MeasurementProvided object when the result is Double" in {
```
miroslavpojer, line 29: Test speak about Double, but works with Big Decimals

miroslavpojer (commented):
- [x] pulled
- [x] built
- [ ] test run
- [ ] code coverage - required merge of conflicting files, there were changes in tests
  - in current test implementation I see not covered paths

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -50,122 +49,124 @@ object AtumMeasure {
...
+        if (result == null) 0 else result
+      case _: StringType =>
         val value =
-          if (collected == null) new java.math.BigDecimal(0)
-          else collected.asInstanceOf[java.math.BigDecimal]
+          if (result == null) new java.math.BigDecimal(0)
+          else result.asInstanceOf[java.math.BigDecimal]
         value.stripTrailingZeros // removes trailing zeros (2001.500000 -> 2001.5, but can introduce scientific notation (600.000 -> 6E+2)
           .toPlainString // converts to normal string (6E+2 -> "600")
       case _ =>
```
benedeki, line 166: For the record, this we might want to revisit and rethink. But not part of this PR. 😉

benedeki (commented):
I think it's ok.
Just decide uf you want to add the hash function on multiple columns to add here or separate ticket+PR.

Oh and there's Miroslav's comment on one of the tests.

salamonpavel:
I will re-point this PR to master instead of feature/99-custom-measure-2.

benedeki (approved):

PR #113, opened 2023-11-07
#23: normalizing Scala -> PG objects

lsulak:
Related to #23, but won't fully implement everything in that ticket. I'll split it into more PRs because #23 is dependant on DB code that is still not fully done&merged.


server/src/main/scala/za/co/absa/atum/web/api/database/Runs.scala
```diff
@@ -34,32 +36,37 @@ class Runs (implicit dBEngine: SlickPgEngine) extends DBSchema{
...
-      .replace("[", "{")
-      .replace("]", "}")
+  private def nestedScalaSeqToPgArray(toConvert: Seq[Seq[String]]): String = {
+    val scalaSeqJsonized = toConvert.map(scalaSeqToPgArray).mkString(",")
+    s"{$scalaSeqJsonized}"
+  }
+
+  private def scalaSeqToPgArray(toConvert: Seq[String]): String = {
+    val scalaArrayAndItsContent: Regex = "^\\[(.*)\\]$".r
+    val scalaSeqJsonized = SerializationUtils.asJson(toConvert)
```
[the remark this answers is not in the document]
lsulak: Explained via a quick chat

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
```diff
@@ -35,33 +37,34 @@ class Runs (implicit dBEngine: SlickPgEngine) extends DBSchema{
...
-      .replace("[", "{")
-      .replace("]", "}")
+  private def nestedScalaSeqToPgArray[T <: AnyRef: ClassTag](toConvert: Seq[Seq[T]]): String = {
+    val scalaSeqJsonized = toConvert.map(scalaSeqToPgArray).mkString(",")
+    "{" + scalaSeqJsonized + "}"
+  }
+
+  private def scalaSeqToPgArray[T <: AnyRef: ClassTag](toConvert: Seq[T]): String = {
+    val scalaSeqJsonized = SerializationUtils.asJson(toConvert)  // this also correctly escapes double quotes
+    "{" + scalaSeqJsonized.substring(1, scalaSeqJsonized.length - 1) + "}"
```
benedeki, line 47: **Minor:**
```suggestion
    val scalaSeqJsonized = SerializationUtils.asJson(toConvert)                  // this also correctly escapes double quotes
    "{" + scalaSeqJsonized.substring(1, scalaSeqJsonized.length - 1) + "}"  //...only the square brackets have to be replaced with curly ones
```

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
```diff
@@ -35,33 +37,34 @@ class Runs (implicit dBEngine: SlickPgEngine) extends DBSchema{
...
-  private def scalaIterableToSQLArray(toConvert: Seq[Any]): String = {
-    SerializationUtils.asJson(toConvert)
-      .replace("[", "{")
-      .replace("]", "}")
+  private def nestedScalaSeqToPgArray[T <: AnyRef: ClassTag](toConvert: Seq[Seq[T]]): String = {
+    val scalaSeqJsonized = toConvert.map(scalaSeqToPgArray).mkString(",")
+    "{" + scalaSeqJsonized + "}"
+  }
+
+  private def scalaSeqToPgArray[T <: AnyRef: ClassTag](toConvert: Seq[T]): String = {
```
benedeki, line 45: Why the type limitation (`<: AnyRef: ClassTag`)?

benedeki (approved):
Maybe there's a generic recursive solution, but this is good and enough:

- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
lsulak: This is the definition of `asJson` function that I'm calling inside: `def asJson[T <: AnyRef](obj: T): String` and this generic type `T` here is also limited, so the 'outer' `T` here in my function cannot be a superset of it. The build was failing

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
lsulak: That's better, thanks, I'll change it :)

benedeki (approved):

Issue #103, opened 2023-10-30
Add DB tests


benedeki:
## Background
There's a lot logic in DB functions, but there are no tests.

## Feature
Add tests of the DB functions.

Depends on #48

benedeki:
This item depends on:
* AbsaOSS/atum-service#48
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

benedeki:
This item depends on:
* ~~AbsaOSS/atum-service#48~~
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

PR #114, opened 2023-11-07
#71: partitioningDTO conversion to JSONB that PG understands 

### LINKED ISSUE #71: Add enhancement of fa-db to convert between `AtumPartitioning` and the composite type
## Background
`AtumPartitioning` is a `LinkedMap[String, String]`, the order of the keys matters.

## Feature
Add enhancement of _fa-db_ to convert between `AtumPartitioning` and the composite type of #68.

Depends on #68

lsulak:
It also and keeps the partitioning valid & key order is kept. No extra checks implemented here - in theory partitioning can be 'empty', i.e. it can have just keys, not values - let the DB decide what it needs - checks will be implemented there.

Closes #71.


server/src/main/scala/za/co/absa/atum/web/model/Partitioning.scala
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.web.model
+
+import za.co.absa.atum.model.dto.PartitionDTO
+
+case class Partitioning (
```
benedeki, line 21: **Minor:**
Would be considering naming the class `DBPartioning` or `PartiioningForDB` to avoid confusion with Agent class

benedeki, line 21: What about making the constructor `private`, to avoid creation of instance with bogus info?

server/src/main/scala/za/co/absa/atum/web/model/Partitioning.scala
```diff
@@ -0,0 +1,36 @@
...
+  version: Int = 1,
+  keys: Seq[String],
+  keysToValuesMap: Map[String, String]
+)
+
+object Partitioning {
+
+  def fromSeqPartitionDTO(partitioning: Seq[PartitionDTO]): Partitioning = {
+    val allKeys = partitioning.map(_.key)
```
benedeki, line 30: **Question:**
Should the keys be case-insensitive maybe? 🤔

server/src/main/scala/za/co/absa/atum/web/model/Partitioning.scala
lsulak: Sure thing. I was even thinking about it and hten forgot 🤦

lsulak: Sure, good idea actually 👍

server/src/main/scala/za/co/absa/atum/server/model/PartitioningForDB.scala
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.server.model
+
+import za.co.absa.atum.model.dto.PartitionDTO
+
+private[server] case class PartitioningForDB private(
```
benedeki, line 21: I meant making the constructor private, didn't need to make the whole class so. But almost as good. 😄

benedeki (commented):
#103: Add DB tests

- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

server/src/main/scala/za/co/absa/atum/server/model/PartitioningForDB.scala
lsulak: It is done

benedeki (approved):

Issue #52, opened 2023-07-17
Call SaveCheckpoint REST API endpoint


benedeki:
## Background
Exchange data between the agent and the server

## Feature
Call the SaveCheckpoint REST API endpoint of the Atum server and provide it with data

Depends on #37
Depends on #38 
Depends on #50

benedeki:
This item depends on:
* AbsaOSS/atum-service#37
* AbsaOSS/atum-service#38
* AbsaOSS/atum-service#50
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

benedeki:
This item depends on:
* ~~AbsaOSS/atum-service#37~~
* ~~AbsaOSS/atum-service#38~~
* ~~AbsaOSS/atum-service#50~~
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

PR #119, opened 2023-11-11
#23: add partitioning fa db call

### LINKED ISSUE #23: Create the FA-DB entities for the DB functions

lsulak:
Closes #23.

Draft: I need to add `byUser` attribute to the client's code as it was just recently added to the DTO and server. 
**Update**: Done, author/byUser field is derived automatically; but I still need to finish the endpoint
**Update2**: Done, everything tested and it seems to work


server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
```diff
@@ -84,4 +86,29 @@ object Runs {
 
     override protected def slickConverter: GetResult[Unit] = GetResult { _ => }
   }
+
+  class CreatePartitioningIfNotExists(implicit override val schema: DBSchema, override val dbEngine: SlickPgEngine)
+    extends DBSingleResultFunction[PartitioningDTO, Long, SlickPgEngine]
```
lsulak, line 91: Note: this is `Long` and not `BigInt` as in the DB, but `Long` in Scala is the same as `BigInt` in PG - 8byte signed integer.

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
```diff
@@ -84,4 +86,29 @@ object Runs {
...
+      with StandardStatusHandling {
+
+    override protected def sql(values: PartitioningDTO): SQLActionBuilder = {
+      val partitioning = PartitioningForDB.fromSeqPartitionDTO(values.partitioning)
+      val partitioningNormalized = SerializationUtils.asJson(partitioning)
+
+      val parentPartitioningNormalized = values.parentPartitioning.map { parentPartitioning => {
+        val parentPartitioningForDB = PartitioningForDB.fromSeqPartitionDTO(parentPartitioning)
+        SerializationUtils.asJson(parentPartitioningForDB)
+      }}.getOrElse("{}")
```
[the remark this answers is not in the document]
lsulak: Ok I changed this, I'll test this later, hopefully today

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
lsulak: I was strictly following your DB function that returns `OUT id_partitioning`. But when I think about it now, we need to retrieve the original author of the partitioning so some sort of SELECT will be needed

lsulak: Regarding ID in the server, or in the client - I don't think that's needed, we can consider all IDs (except for Checkpoint ID) to be something that only DB is aware of, other pieces of the app don't care

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -45,25 +56,36 @@ class AtumAgent private[agent] () {
...
+    val authorIfNew = AtumAgent.currentUser
     val newPartitions: AtumPartitions = parentAtumContext.atumPartitions ++ subPartitions
 
     val newPartitionsDTO = AtumPartitions.toSeqPartitionDTO(newPartitions)
     val parentPartitionsDTO = Some(AtumPartitions.toSeqPartitionDTO(parentAtumContext.atumPartitions))
-    val partitioningDTO = PartitioningDTO(newPartitionsDTO, parentPartitionsDTO)
+    val partitioningDTO = PartitioningSubmitDTO(newPartitionsDTO, parentPartitionsDTO, authorIfNew)
 
     val atumContextDTO = dispatcher.getOrCreateAtumContext(partitioningDTO)
     lazy val atumContext = AtumContext.fromDTO(atumContextDTO, this)
```
benedeki, line 87: Why is this lazy @salamonpavel ?
If it's do avoid unnecessary creation, it would be enough to send directly `AtumContext.fromDTO(atumContextDTO, this)` into `getExistingOrNewContext`.
`lazy` has too many other undesired consequences.

salamonpavel: I remember I was asked to refactor the AtumContext.fromDTO into a val before passing into the function call. I agree with the suggested improvement. In this particular case which undesired consequences are realistic in your opinion?

database/src/main/postgres/runs/write_checkpoint.sql
```diff
@@ -92,9 +92,9 @@ BEGIN
     PERFORM runs._write_measurement(
         i_id_checkpoint,
         _fk_partitioning,
-        UN.measurement->>'measureName',
-        jsonb_array_to_text_array(UN.measurement->'measuredColumns'),
-        UN.measurement->'measurementValue',
+        UN.measurement->'measure'->>'measureName',
+        jsonb_array_to_text_array(UN.measurement->'measure'->'measuredColumns'),
+        UN.measurement->'result'->'mainValue',
```
benedeki, line 97:
```suggestion
        UN.measurement->'result',
```
We want to have the whole result there.

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -33,6 +35,15 @@ class AtumAgent private[agent] () {
     case dt => throw new UnsupportedOperationException(s"Unsupported dispatcher type: '$dt''")
   }
 
+  /**
+   * Returns a user under who's security context the JVM is running. It can be used for author/createdBy fields.
```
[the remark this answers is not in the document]
lsulak: Agree. That's better, thanks

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -18,13 +18,15 @@ package za.co.absa.atum.agent
...
 import za.co.absa.atum.agent.dispatcher.{ConsoleDispatcher, HttpDispatcher}
-import za.co.absa.atum.model.dto.{CheckpointDTO, PartitioningDTO}
+import za.co.absa.atum.model.dto.{CheckpointDTO, PartitioningSubmitDTO}
 
 /**
  * Entity that communicate with the API, primarily focused on spawning Atum Context(s).
  */
 class AtumAgent private[agent] () {
 
+  private[this] var contexts: Map[AtumPartitions, AtumContext] = Map.empty
```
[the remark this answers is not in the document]
lsulak: Because having such an important attribute in the bottom of this class, where it's almost invisible, is really weird. 

Having vals/vars on the bottom, and methods first, is odd habit that leads to harder to read code in my personal opinion. I've never saw this in the past in any programming language. Even Java has some recommendations and it's not like this, on SO it's being mentioned: https://softwareengineering.stackexchange.com/questions/380771/in-which-order-functions-should-be-defined-scala-methods. This is usually never enforced and not a 'best practice', but methods are usually the last ones in a class, class variables first

database/src/main/postgres/runs/write_checkpoint.sql
lsulak: You are absolutely right, will change it asap

server/src/main/scala/za/co/absa/atum/server/api/service/DatabaseService.scala
```diff
@@ -16,36 +16,30 @@
...
-  /**
-   * Function to retrieve checkpoint based on the provided fields
-   *
-   * @param filterCriteria JSON object containing the fields for filtering the checkpoint
-   */
-  def createPartitioningIfNotExists(partitioningInfo: PartitioningDTO): Seq[PartitionDTO] = {
-    // Todo - implement the db function call in #23
-    Seq()
+  def createPartitioningIfNotExists(partitioning: PartitioningSubmitDTO): CompletableFuture[PartitioningSubmitDTO] = {
+    implicit val executionContext: ExecutionContext = postgresAccessProvider.executor
```
benedeki, line 40: I would not use the implicit this way. Either make it on the class level, or hand it over explicitly.

server/src/main/scala/za/co/absa/atum/server/api/service/DatabaseService.scala
```diff
@@ -16,36 +16,30 @@
...
 
-  /**
-   * Function to retrieve checkpoint based on the provided fields
-   *
-   * @param filterCriteria JSON object containing the fields for filtering the checkpoint
-   */
-  def createPartitioningIfNotExists(partitioningInfo: PartitioningDTO): Seq[PartitionDTO] = {
-    // Todo - implement the db function call in #23
-    Seq()
+  def createPartitioningIfNotExists(partitioning: PartitioningSubmitDTO): CompletableFuture[PartitioningSubmitDTO] = {
```
benedeki, line 39: **To consider:**
Would make all the classes work with Scala's `Future` only the controller then Java's `CompletableFuture`.
I want to avoid the mess UU turned into.

server/src/main/scala/za/co/absa/atum/server/api/service/DatabaseService.scala
lsulak: Okay

server/src/main/scala/za/co/absa/atum/server/api/controller/PrototypeController.scala
```diff
@@ -41,19 +43,26 @@ class PrototypeController @Autowired()(databaseService: DatabaseService){
...
   @PostMapping(Array("/createPartitioning"))
   @ResponseStatus(HttpStatus.OK)
-  def createPartitioningIfNotExists(@RequestBody partitioningInfo: PartitioningDTO): AtumContextDTO = {
-    val partitioning = databaseService.createPartitioningIfNotExists(partitioningInfo)
-    val measures = Set.empty[MeasureDTO]
+  def createPartitioning(@RequestBody partitioning: PartitioningSubmitDTO): CompletableFuture[AtumContextDTO] = {
+    val partitioningFuture = databaseService.createPartitioningIfNotExists(partitioning)
+    val measures: Set[MeasureDTO] = Set(
+      // TODO #120, get measures from DB, this solution is temporary - we need record count always for now
+      MeasureDTO("count", Seq("countCol"))
```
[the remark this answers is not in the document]
lsulak: Change on the way, sure

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
```diff
@@ -72,13 +73,13 @@ class AtumContext private[agent] (
     this
   }
 
-  def createCheckpointOnProvidedData(checkpointName: String, author: String, measurements: Seq[Measurement]): AtumContext = {
+  def createCheckpointOnProvidedData(checkpointName: String, measurements: Seq[Measurement]): AtumContext = {
     val offsetDateTimeNow = OffsetDateTime.now()
 
     val checkpointDTO = CheckpointDTO(
```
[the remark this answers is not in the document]
lsulak: Hmm, like a companion object for this DTO that would have an apply method?

server/src/main/scala/za/co/absa/atum/server/api/service/DatabaseService.scala
lsulak: Okay

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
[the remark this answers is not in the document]
benedeki: Actually you are right, that would be a wrong direction dependency.. Maybe instead a method in Portioning to create the DTO? The thing is, DTO is created twice the same way - DRY.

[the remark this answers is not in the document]
lsulak: Yep exactly. I wanted to have DTOs that are a simple case classes, no logic or so around them. Okay aI'll think about it :)

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
```diff
@@ -30,13 +30,14 @@ class HttpDispatcher(config: Config) extends Dispatcher with Logging {
...
 
-  override def getOrCreateAtumContext(partitioning: PartitioningDTO): AtumContextDTO = {
+  override def getOrCreateAtumContext(partitioning: PartitioningSubmitDTO): AtumContextDTO = {
     basicRequest
       .body(s"$partitioning")
       .post(serverUri)
       .send(backend)
 
-    AtumContextDTO(partitioning = partitioning.partitioning) // todo: implement request
+    // todo: implement request
```
miroslavpojer, line 39: Fotgotten TODO? Or planned in following PR?

server/src/main/scala/za/co/absa/atum/server/api/controller/PrototypeController.scala
```diff
@@ -42,18 +42,26 @@ class PrototypeController @Autowired()(databaseService: DatabaseService){
...
+    )
     val additionalData = AdditionalDataDTO(additionalData = Map.empty)
 
-    AtumContextDTO(partitioning, measures, additionalData)
+    partitioningFuture
+      .thenApply(
+        partitioningObtained => AtumContextDTO(partitioningObtained.partitioning, measures, additionalData)
+      )
   }
+
```
miroslavpojer, line 66: Why empty line?

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
```diff
@@ -33,6 +33,7 @@ class Runs (implicit dBEngine: SlickPgEngine) extends DBSchema{
   import Runs._
 
   val writeCheckpoint = new WriteCheckpoint
+  val createPartitioningIfNotExists = new CreatePartitioningIfNotExists
 }
 
 object Runs {
```
miroslavpojer, line 39: Method `nestedScalaSeqToPgArray` is not more used.

miroslavpojer (commented):
- [x] pulled
- [x] code reviewed
- [x] run tests
- [ ] check code coverage - No JaCoCo report present as target branch is not master.

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
lsulak: Yeah neither is the one below, but I think it will be, once we'll use TEXT[] or TEXT[][] fields.

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
lsulak: Not forgotten, @benedeki  is gonna work on it soon: https://github.com/AbsaOSS/atum-service/issues/52

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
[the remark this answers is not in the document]
lsulak: Actually...is it really worth it? It's on 2 places that are almost the same, yes, but is it such a big issue? It will probably not grow.  

Such method / function you propose would have tons of parameters and I don't think it would really make the code better. If you insist I can, but I'm just not convinced to create auxiliary functions when the internals of such functions are so trivial & when it's not reusable on more places

server/src/main/scala/za/co/absa/atum/server/api/controller/PrototypeController.scala
lsulak: probably just to have an 'empty space' between individual endpoints, before and after them

server/src/main/scala/za/co/absa/atum/server/api/database/Runs.scala
benedeki: I don't think we will ever use `TEXT[][]` becuse of the subarrays` limitation (all the same size).
This one can be safely removed.

lsulak: done

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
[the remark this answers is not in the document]
benedeki: Interesting I was thought and number of my mentors was of the same opinion that in front are the public member of the class, then protected and private ones at the end.
Anyway, not worth further discussion, definitely not now 😉

agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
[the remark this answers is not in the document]
benedeki: Would be nice, but not a priority now. 👍

server/src/main/scala/za/co/absa/atum/server/api/controller/PrototypeController.scala
lsulak: matter of personal style I would say 🤔

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

PR #130, opened 2023-11-20
wrong format of application.properties file

lsulak:
Will need to retrigger the ECR push once this is merged. But not sure if this really is causing the problems.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -53,12 +53,12 @@ ARG SCALA_VERSION=2.12
 LABEL org.opencontainers.image.authors="ABSA"
 
 # Copy Spring application properties
-COPY $CONFIG /opt/application.yaml
+COPY $CONFIG /opt/application.properties
 
 # deploy as root application in tomcat
 COPY ${AS_PREFIX}/target/jvm-${SCALA_VERSION}/*.war /usr/local/tomcat/webapps/ROOT.war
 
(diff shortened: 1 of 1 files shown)


jakipatryk (approved):

server/Dockerfile
```diff
@@ -53,12 +53,12 @@ ARG SCALA_VERSION=2.12
...
 
 # Copy Spring application properties
-COPY $CONFIG /opt/application.yaml
+COPY $CONFIG /opt/application.properties
 
 # deploy as root application in tomcat
 COPY ${AS_PREFIX}/target/jvm-${SCALA_VERSION}/*.war /usr/local/tomcat/webapps/ROOT.war
 
-ENV SPRING_CONFIG_LOCATION=/opt/application.yaml
+ENV SPRING_CONFIG_LOCATION=/opt/application.properties
```
benedeki, line 61: Is this the correct location?

benedeki (commented):
How are the actual values supplied into the `application.conf` file?

lsulak:
> How are the actual values supplied into the `application.conf` file?

a brand new application.properties file is crafted and its path is supplied to the `SPRING_CONFIG_LOCATION`

server/Dockerfile
lsulak: @Zejnilovic used the same approach as in Login Service

benedeki (approved):

Issue #107, opened 2023-11-01
Server -> DB: Add pulling DB login credentials via AWS Secrets


lsulak:
## Background
Please have inspiration in https://github.com/absa-group/AUL/pull/1907.

PR #135, opened 2023-11-29
Server db add pulling db login credentials via aws secrets:#107

TebaleloS:
Added a singleton class to retrieve secret key from aws


project/Dependencies.scala
```diff
@@ -137,8 +143,16 @@ object Dependencies {
     lazy val faDbSlick = "za.co.absa.fa-db" %% "slick" % Versions.fadb
     lazy val slickPg = "com.github.tminglei" %% "slick-pg" % Versions.slickpg
 
+    // awssdk library for providing the functionality of retrieving the secret key
+    lazy val awsSdk = "software.amazon.awssdk" % "secretsmanager" % Versions.awssdk
+    lazy val sso = "software.amazon.awssdk" % "sso" % Versions._sso
+
+    // For allowing array of strings to be parsed as json
+    lazy val spray = "io.spray" %% "spray-json" % Versions.spray
```
salamonpavel, line 151: What's the reason to introduce yet another json library into the project?

TebaleloS: That will be removed. It was before I realize that I didn't even need convert the response to json.

server/src/main/scala/za/co/absa/atum/server/api/RetrieveAwsSecret.scala
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      val request = GetSecretValueRequest.builder()
+        .secretId(secretName)
+        .build()
+
+      val response = secretsManagerClient.getSecretValue(request)
+      response.secretString.map(_.toString)
+    } catch {
+      case e: SecretsManagerException =>
+        println(s"Error retrieving secret key: ${e.getMessage}")
```
salamonpavel, line 50: You might consider using logger instead of println.

server/src/main/scala/za/co/absa/atum/server/api/RetrieveAwsSecret.scala
```diff
@@ -0,0 +1,58 @@
...
+import software.amazon.awssdk.auth.credentials.ProfileCredentialsProvider
+import software.amazon.awssdk.regions.Region
+import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient
+import software.amazon.awssdk.services.secretsmanager.model.{GetSecretValueRequest, SecretsManagerException}
+
+
+/**
+ * Class implement the functionality of retrieving secret keys from aws secret manger service
+ */
+class RetrieveAwsSecret (profileCredentials: String = "npintdebdtools-sso") {
```
salamonpavel, line 29: What is the reason for the default value here?

TebaleloS: My bad, I forgot to remove it. I had that with the first test cases that I had. I'm removing it now.

server/src/main/scala/za/co/absa/atum/server/api/provider/PostgresAccessProvider.scala
```diff
@@ -19,28 +19,85 @@ package za.co.absa.atum.server.api.provider
...
+      .addSecretName(secretsConfig, portNumber)
+      .addSecretName(secretsConfig, databaseName)
+      .addSecretName(secretsConfig, user)
+      .addSecretName(secretsConfig, password)
+  }
+
+  private def overrideWithSecret(oldConfig: Config, path: String, secretName: String): Config = {
+
+    val secretString: Seq[String] = retrieveAwsSecret.retrieveAwsSecret(secretName)
+    val overrideValue = secretString.foldLeft(Try("")) {(acc, s) =>
```
salamonpavel, line 65: What's the reason for using Try here? What are you trying to accomplish here, please explain.

server/src/main/scala/za/co/absa/atum/server/api/RetrieveAwsSecret.scala
TebaleloS: In fact, I wanted to put "default" as a value for that. I think it will be a good idea for those who will be using default profiles.

server/src/main/scala/za/co/absa/atum/server/api/provider/PostgresAccessProvider.scala
TebaleloS: So that it can catch the exception in case it happens during the of string concatenation operation.

server/src/main/scala/za/co/absa/atum/server/api/RetrieveAwsSecret.scala
```diff
@@ -0,0 +1,61 @@
...
+  def retrieveAwsSecret(secretName: String): Seq[String] = {
+    val logger: Logger = LoggerFactory.getLogger(getClass.getName)
+
+    try {
+      val request = GetSecretValueRequest.builder()
+        .secretId(secretName)
+        .build()
+
+      val response = secretsManagerClient.getSecretValue(request)
+      response.secretString.map(_.toString)
```
salamonpavel, line 50: @TebaleloS 
I have doubts about this line. `response.secretString` returns String so you are basically mapping every Char of that String into a String. I am confused ... why that has to be done? Isn't there only one secret value for a given secret key/name?

server/src/main/scala/za/co/absa/atum/server/api/RetrieveAwsSecret.scala
```diff
@@ -0,0 +1,61 @@
...
+    } catch {
+      case e: SecretsManagerException =>
+        logger.error(s"Error retrieving secret key: ${e.getMessage}")
+        e.getMessage.map(_.toString)
```
salamonpavel, line 54: This is certainly incorrect. You are again mapping over a `String`, therefore iterating over it's individual `Char` elements. Moreover you don't want to return this as a result of the method. How can a description of an exception `e.getMessage` be a valid return value of this method?

server/src/main/scala/za/co/absa/atum/server/api/RetrieveAwsSecret.scala
TebaleloS: Happy to have a pair coding so that we can do it proper.

server/src/main/scala/za/co/absa/atum/server/api/RetrieveAwsSecret.scala
TebaleloS: Thanks for bringing that into my attention, I actually cooked that so it can silence the compiler. I have refactored it.

server/src/test/scala/za/co/absa/atum/server/api/RetrieveAwsSecretTest.scala
```diff
@@ -0,0 +1,67 @@
...
+      override val secretsManagerClient: SecretsManagerClient = mockSecretsManagerClient
+    }
+    val testSecretName: String = "testSecret"
+    val expectedResults = Success("testSecret")
+
+    val mockResponse = GetSecretValueResponse.builder().secretString(testSecretName).build()
+    when(mockSecretsManagerClient.getSecretValue(any[GetSecretValueRequest])).thenReturn(mockResponse)
+
+    val actualResults = retrieveAwsSecret.retrieveAwsSecret(testSecretName)
+    val overrideValue = actualResults.foldLeft(Try("")) { (acc, s) =>
```
salamonpavel, line 42: Here you foldLeft over a actualResults which is of type String. There is no reason to do such operation.

salamonpavel:
Obsolete. The issue has been addressed in #153.

PR #142, opened 2023-12-20
#117: moving future/obsolete server code elsewhere, wanting to keep the current & actual server module simple

### LINKED ISSUE #117: Server: move obsolete/future code elsewhere
## Background
Currently there is quite a lot of code under `server/` module, but a lot of it is unused. The goal of this ticket is to move such scala files into a dedicated directory (or so), in a similar fashion as what was done on DB side (there is a dir called `future/`.

Please, as part of this exercise, check if we will really be using everything - delete what doesn't make sense anymore, we don't want to have a dead code in our repository. We can always find it via git history if absolutely needed.

lsulak:
Closes #117


benedeki (approved):
Just checked the code (in speed).

PR #150, opened 2024-01-22
introduce zio into atum server

### LINKED ISSUE #147: Replace Spring by ZIO with Http4s and Tapir in Atum server
Introduce (replace Spring by) ZIO with Http4s and Tapir in Atum server.

### LINKED ISSUE #148: Use Doobie module classes instead of Slick
Since we are operating in the context of effectful ZIO types it would be good to replace Slick based functions for Doobie based ones in Atum server module.

### LINKED ISSUE #149: Build Atum server for 2.13
Once the ticket #148 is done we will be able to build Atum server for Scala 2.13.

salamonpavel:
ZIO framework was introduced to Atum server module as a replacement of Spring. Endpoints were defined using Tapir since available interpreters can generate Http routes for Http4s server backend and Swagger docs from Tapir endpoints. Simple database integration tests were introduced to verify that Doobie based function objects work as expected as they replaced Slick based ones. Atum server module's target Scala version upgraded to 2.13. Jacoco plugin upgraded to 1.6.1 as the previous version wasn't compatible with upgraded github runners anymore. Missing test coverage partially addressed. 

! Classes currently not covered by test cases (mainly endpoints) will be tested as part of #152 when we upgrade target Java version #151 !

Closes #147 
Closes #148 
Closes #149


database/src/main/postgres/public/V1.2.2__jsonb_array_to_text_array.sql
```diff
@@ -30,10 +29,10 @@ CREATE OR REPLACE FUNCTION public.jsonb_array_to_text_array(
...
-    LANGUAGE sql IMMUTABLE STRICT PARALLEL SAFE
-BEGIN ATOMIC
-    SELECT ARRAY(SELECT jsonb_array_elements_text(i_json_array));
-END;
-
+LANGUAGE sql IMMUTABLE STRICT PARALLEL SAFE
+AS
+    $$
+        SELECT ARRAY(SELECT jsonb_array_elements_text(i_json_array));
+    $$;
```
lsulak, line 36: If I understood it correctly, I would keep the original actually:
* https://dba.stackexchange.com/questions/300507/what-does-begin-atomic-mean-in-a-postgresql-sql-function-procedure/300512#300512
* https://stackoverflow.com/questions/12144284/what-are-used-for-in-pl-pgsql/12172353#12172353
* see the `sql_body` segment: https://www.postgresql.org/docs/14/sql-createfunction.html

Basically, this is a SQL function and this `BEGIN ATOMIC` is new syntax available from PG 14+ and it's doing parsing at function definition time, not runtime as with your changes.

This is similar to writing the text of the function body as a string constant (see definition above), but there are some differences: This form only works for LANGUAGE SQL, the string constant form works for all languages. This form is parsed at function definition time, the string constant form is parsed at execution time; therefore this form cannot support polymorphic argument types and other constructs that are not resolvable at function definition time.

database/src/main/postgres/public/V1.2.2__jsonb_array_to_text_array.sql
```diff
@@ -30,10 +29,10 @@ CREATE OR REPLACE FUNCTION public.jsonb_array_to_text_array(
...
 
-GRANT EXECUTE ON FUNCTION public.jsonb_array_to_text_array(JSONB) TO public;
+GRANT EXECUTE ON FUNCTION public.jsonb_array_to_text_array(JSONB) TO public;
```
lsulak, line 38: no newline at the end of file

server/src/main/scala/za/co/absa/atum/server/Main.scala
```diff
@@ -0,0 +1,52 @@
...
+    server
+      .provide(
+        PartitioningControllerImpl.layer,
+        CheckpointControllerImpl.layer,
+        PartitioningServiceImpl.layer,
+        CheckpointServiceImpl.layer,
+        PartitioningRepositoryImpl.layer,
+        CheckpointRepositoryImpl.layer,
+        CreatePartitioningIfNotExists.layer,
+        WriteCheckpoint.layer,
```
lsulak, line 43: it's interesting (to me) that all this is added here. Combining controller implementations with DB functions and repositories - but I guess that the reason is automatic dependency injection, i.e. `WriteCheckpoint` into `CheckpointRepositoryImpl` (really just a wild guess)

salamonpavel: Yeah, basically you can provide all layers and the framework will figure out the dependency graph for you. Obviously if you feel we should construct the final layer some place else in the code base it's possible.

database/src/main/postgres/public/V1.2.2__jsonb_array_to_text_array.sql
salamonpavel: Ok, will add it, once done I will update this comment.

database/src/main/postgres/public/V1.2.2__jsonb_array_to_text_array.sql
salamonpavel: Let me explain. I have changed it to this version so that I can use the Flyway plugin. I can see that it might have not been the best idea, after all I could keep the original one and add use this one just for the purposes of local testing. The thing is that Flyway works only with Postgres up to version 13. It fails with the original version of the function.

Would that be better if I keep both for now? By the way I hope we will soon be able to somehow automate this db stuff with Ultet. If I understood correctly it should be able to do such things.

server/src/main/scala/za/co/absa/atum/server/Constants.scala
```diff
@@ -14,11 +14,13 @@
...
-  controlCol: String,
-  controlValue: Any
-)
+object Constants {
+  final val Api = "api"
+  final val V1 = "v1"
+  final val CreatePartitioning = "createPartitioning"
+  final val CreateCheckpoint = "createCheckpoint"
+  final val SwaggerApiName = "Atum API"
+  final val SwaggerApiVersion = "1.0"
```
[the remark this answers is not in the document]
salamonpavel: Upper camel case is a convention in Scala for constants. Below some examples.
```scala
Int.MaxValue
scala.math.Pi
```
https://docs.scala-lang.org/style/naming-conventions.html

server/src/main/resources/reference.conf
```diff
@@ -0,0 +1,23 @@
...
+         # rootLevel sets the minimum level of log messages that will be displayed
+         rootLevel = INFO
+      }
+   }
+   postgres {
+      # The JDBC driver class
+      dataSourceClass=org.postgresql.Driver
+      serverName=localhost
+      # set to non-default port, to avoid local collision
+      portNumber=5432
```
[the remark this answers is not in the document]
salamonpavel: Yeah, will fix that, initially I used 5433, now it's set for the default port.

server/src/main/resources/reference.conf
```diff
@@ -0,0 +1,23 @@
...
+      databaseName=atum_db
+      user=atum_user
+      password=changeme
+      # maximum number of connections that HikariCP will keep in the pool, including both idle and in-use connections
```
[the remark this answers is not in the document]
salamonpavel: In Doobie the connection is established differently than in Slick. Notice that I use `HikariTransactor` which works with HikariCP. We don't need to pass it as a parameter in the configuration anymore.
```scala
object TransactorProvider {

  val layer: ZLayer[Any with Scope, Throwable, HikariTransactor[Task]] = ZLayer {
    for {
      postgresConfig <- ZIO.config[PostgresConfig](PostgresConfig.config)
      hikariConfig = {
        val config = new HikariConfig()
        config.setDriverClassName(postgresConfig.dataSourceClass)
        config.setJdbcUrl(
          s"jdbc:postgresql://${postgresConfig.serverName}:${postgresConfig.portNumber}/${postgresConfig.databaseName}"
        )
        config.setUsername(postgresConfig.user)
        config.setPassword(postgresConfig.password)
        config.setMaximumPoolSize(postgresConfig.maxPoolSize)
        config
      }
      xa <- HikariTransactor
        .fromHikariConfig[Task](hikariConfig, defaultBlockingExecutor.asExecutionContext)
        .toScopedZIO
    } yield xa
  }

}
```

database/README.md
```diff
@@ -0,0 +1,13 @@
+## How to set up database for local testing
```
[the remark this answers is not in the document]
salamonpavel: Yeah I hope we will solve this DB stuff once and for all in order to enable simple database testing. One shouldn't spend unreasonable amount of time by setting up the database or by cleaning it up. Flyway unfortunately doesn't support latest versions of Postgres so please think of all the things I did in this regard as temporary.

build.sbt
```diff
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-import Dependencies._
-import SparkVersionAxis._
-import JacocoSetup._
+import Dependencies.*
+import SparkVersionAxis.*
+import JacocoSetup.*
```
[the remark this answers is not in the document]
salamonpavel: Actually I didn't change this. I have no idea how that happened. And it probably doesn't matter since it still imports everything. I know that `*` symbol is used in Scala 2.13 and later in many places where previously `_` was used.

.github/workflows/build.yml
```diff
@@ -37,9 +51,14 @@ jobs:
...
         uses: actions/checkout@v2
       - uses: coursier/cache-action@v5
+
       - name: Setup Scala
         uses: olafurpg/setup-scala@v10
         with:
           java-version: "adopt@1.8"
+
+      - name: Prepare testing database
+        run: sbt flywayMigrate
```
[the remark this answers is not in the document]
salamonpavel: You've hit the nail on the head and I couldn't agree more. On the other hand every project I've worked on so far didn't have tests for database functions. This has been a bit of a wake-up call for me. 

I've seen firsthand how tricky it can be to refactor code that hasn't been tested. So, I've decided to take a different route. I'm committed to making sure my code is tested, no matter what.

As for Flyway, I see it as a stepping stone. It's a good temporary fix as we plan on switching to Ultet when it's released.

.github/workflows/jacoco_check.yml
```diff
@@ -45,11 +45,11 @@ jobs:
         with:
           java-version: "adopt@1.8"
       - name: Build and run tests
-        run: sbt ++${{matrix.scala}} jacoco
+        run: sbt ++${{matrix.scala}} "agent/jacoco"
```
miroslavpojer, line 48: This line produces only build for 2.13. So there are missing XML report values.

miroslavpojer (commented):
- [ ] pulled
- [ ] jacoco review.

server/src/main/scala/za/co/absa/atum/server/api/database/PostgresDatabaseProvider.scala
```diff
@@ -14,16 +14,20 @@
...
-class TestService @Autowired()(atumConfig: AtumConfig){
-  def getMessage: String = {
-    val testConfigVal = atumConfig.testEndpointConfig.getString("some-key")
-    s"The service says: alfa '$testConfigVal'"
+class PostgresDatabaseProvider(val dbEngine: DoobieEngine[Task])
+
+object PostgresDatabaseProvider {
+  val layer: URLayer[Transactor[Task], PostgresDatabaseProvider] = ZLayer {
+    for {
+      transactor <- ZIO.service[Transactor[Task]]
```
lsulak, line 29: is `TransactorProvider` somehow gonna be injected here? Because `TransactorProvider` is not used anywhere apart from being imported into `Main.scala` as a layer

server/src/main/scala/za/co/absa/atum/server/api/database/runs/functions/WriteCheckpoint.scala
```diff
@@ -0,0 +1,74 @@
...
+import za.co.absa.fadb.doobie.{DoobieEngine, StatusWithData}
+import za.co.absa.fadb.doobie.DoobieFunction.DoobieSingleResultFunctionWithStatus
+import za.co.absa.fadb.status.handling.implementations.StandardStatusHandling
+import za.co.absa.atum.server.api.database.PostgresDatabaseProvider
+import za.co.absa.atum.server.api.database.runs.Runs
+
+import zio._
+import zio.interop.catz._
+
+import doobie.postgres.implicits._
```
lsulak, line 35: the intelliJ says that this is not used and I couldn't find the usages either. Is it just nice to have in case that we'll add something?

database/src/main/postgres/flow_patterns/V1.4.1___.ddl
lsulak, line 1: why there are sometimes 3x `_` and sometimes 2x `_` for these SQL files?

server/src/main/scala/za/co/absa/atum/server/api/http/Server.scala
```diff
@@ -0,0 +1,101 @@
...
+  private def createAllServerRoutes: HttpRoutes[F] = {
+    val endpoints = List(
+      createServerEndpoint(createCheckpointEndpoint, CheckpointController.createCheckpoint),
+      createServerEndpoint(createPartitioningEndpoint, PartitioningController.createPartitioningIfNotExists)
+    )
+    ZHttp4sServerInterpreter[Env](http4sServerOptions).from(endpoints).toRoutes
+  }
+
+  private def createSwaggerRoutes: HttpRoutes[F] = {
+    val endpoints = List(createCheckpointEndpoint, createPartitioningEndpoint)
```
lsulak, line 85: perhaps consider extracting these elsewhere so that it's automatically loaded. I.e., whoever adds more endpoints must add such newly created classes/objects to multiple places, and here I propose to implement it in a way that keep such places to minimum

server/src/test/scala/za/co/absa/atum/server/api/controller/CheckpointControllerSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+        test("Returns expected CheckpointDTO") {
+          for {
+            result <- CheckpointController.createCheckpoint(checkpointDTO1)
+          } yield assertTrue(result == checkpointDTO1)
+        },
+        test("Returns expected InternalServerErrorResponse") {
+          assertZIO(CheckpointController.createCheckpoint(checkpointDTO3).exit)(failsWithA[InternalServerErrorResponse])
+        },
+        test("Returns expected GeneralErrorResponse") {
+          assertZIO(CheckpointController.createCheckpoint(checkpointDTO2).exit)(failsWithA[GeneralErrorResponse])
```
lsulak, line 58: todo for me, I want to ask about this

server/src/test/scala/za/co/absa/atum/server/api/http/EndpointsSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//package za.co.absa.atum.server.api.http
+//
+//import org.scalatest.funspec.AnyFunSpec
+//import org.scalatest.matchers.should.Matchers
```
lsulak, line 20: basically dead code for the whole module

lsulak: still todo, or?

server/src/test/scala/za/co/absa/atum/server/api/TestTransactorProvider.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import doobie.HC
+import doobie.util.transactor.{Strategy, Transactor}
+import za.co.absa.atum.server.config.PostgresConfig
+import zio._
+import zio.interop.catz._
+
+object TestTransactorProvider {
+
+  val layer: ZLayer[Any, Config.Error, Transactor[Task]] = ZLayer {
```
lsulak, line 27: I really like that you implemented both layers like this. I would consider further distinguishing between those two by renaming this one to `layerWithoutRollback` perhaps

lsulak: (also todo for me, I want to ask about some details related to this and differences to the TransactorProvider that lives outside tests)

database/src/main/postgres/public/V1.2.2__jsonb_array_to_text_array.sql
lsulak: Okay, understood, in that case..given how trivial this is & it's not like there are many dependencies for this, I guess it's okay if the parsing is not during function definition time. We can create a ticket for this one to get it back.

build.sbt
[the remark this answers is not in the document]
lsulak: Since we still have to generally support Scala 2.12 I would rather move it back to `_` - https://github.com/scala/scala/releases/tag/v2.13.7 it looks like it's more like a Scala 3 thing, with support in this 2.13 version

.github/workflows/build.yml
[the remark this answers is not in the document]
lsulak: Understood and I really love such approach, it's just that it might be more prudent to create another PR, perhaps pointing to this one, so that it's a bit smaller. But I still really appreciate this!! Just please let's notify the rest of the team about this decision however temporary it might be, it still needs to be communicated in my opinion - I'm not sure about individual capacities for our team, i.e. I don't know who and how detailed reviews of this PR will be done

server/src/test/scala/za/co/absa/atum/server/api/http/EndpointsSpec.scala
lsulak: let's remove this file and create a new issue for Java 8 -> Java 11, and let's put this file there

server/src/main/scala/za/co/absa/atum/server/api/database/DoobieImplicits.scala
```diff
@@ -0,0 +1,156 @@
...
+
+import doobie.{Get, Put}
+import doobie.postgres.implicits._
+
+import org.postgresql.jdbc.PgArray
+import org.postgresql.util.PGobject
+
+import scala.util.{Failure, Success, Try}
+
+object DoobieImplicits {
```
lsulak, line 30: todo I just want to ask about some things related to this module

build.sbt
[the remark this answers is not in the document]
salamonpavel: ok, changed

build.sbt
```diff
@@ -50,6 +50,14 @@ val mergeStrategy: Def.SettingsDefinition = assembly / assemblyMergeStrategy :=
   case _                       => MergeStrategy.first
 }
 
+enablePlugins(FlywayPlugin)
+flywayUrl := "jdbc:postgresql://localhost:5432/atum_db"
+flywayUser := "postgres"
```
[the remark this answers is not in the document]
salamonpavel: I've tried but it seems that these have to be defined in the build.sbt.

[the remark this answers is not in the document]
salamonpavel: I have extracted the config values into a separate file.

database/src/main/postgres/public/V1.2.2__jsonb_array_to_text_array.sql
salamonpavel: I have added the file back in its original form. And added additional one just for the local testing.

database/src/main/postgres/public/V1.2.2__jsonb_array_to_text_array.sql
salamonpavel: done

.github/workflows/build.yml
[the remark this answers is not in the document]
salamonpavel: OK will mention this during our team's meeting.

.github/workflows/jacoco_check.yml
salamonpavel: @miroslavpojer 
Do you happen to know then how to trigger it correctly only for agent module?

server/src/main/scala/za/co/absa/atum/server/api/database/PostgresDatabaseProvider.scala
salamonpavel: As discussed on the call. `Transactor[Task]` will be looked up from zio environment. `TransactorProvided` is only an object where the layer is defined.

server/src/main/scala/za/co/absa/atum/server/api/database/runs/functions/WriteCheckpoint.scala
salamonpavel: It's needed as it imports stuff for UUID and ZonedDataTime, in this particular case. I have no idea why is IntelliJ confused.

server/src/main/scala/za/co/absa/atum/server/api/http/Server.scala
salamonpavel: At this point I can't think of a way how to reduce it. `createAllServerRoutes` creates Http4s server routes and you need to map each endpoint to its controller.

server/src/test/scala/za/co/absa/atum/server/api/http/EndpointsSpec.scala
salamonpavel: I have removed the dependency and also the test code. Created an issue for addressing tests of http endpoints.
https://github.com/AbsaOSS/atum-service/issues/152

server/src/test/scala/za/co/absa/atum/server/api/TestTransactorProvider.scala
salamonpavel: Ok, I have renamed it as per your suggestion.

.github/workflows/jacoco_check.yml
salamonpavel: changed to 
```sh 
sbt "++${{matrix.scala}}; project agent; jacoco"
```

project/FlywayConfiguration.scala
```diff
@@ -1,3 +1,5 @@
+import io.github.davidmweber.FlywayPlugin.autoImport.{flywayLocations, flywayPassword, flywaySqlMigrationSuffixes, flywayUrl, flywayUser}
```
lsulak, line 1: I think that the licence must be on the top of the file

project/Dependencies.scala
```diff
@@ -115,42 +123,57 @@ object Dependencies {
...
+    lazy val playJson = playOrg %% "play-json" % Versions.playJson
+
+    // Fa-db
+    lazy val faDbDoobie = faDbOrg %% "doobie" % Versions.fadb
+
+    // testing
+    lazy val zioTest = zioOrg %% "zio-test" % Versions.zio % Test
+    lazy val zioTestSbt = zioOrg %% "zio-test-sbt" % Versions.zio % Test
+    lazy val zioTestJunit = zioOrg %% "zio-test-junit" % Versions.zio % Test
+    lazy val sbtJunitInterface = "com.github.sbt" % "junit-interface" % "0.13.3" % Test
```
lsulak, line 158: this version here can be extracted to the `object Versions`

project/FlywayConfiguration.scala
```diff
@@ -14,18 +16,12 @@
...
 
-case class ControlMeasure(
-  id: Option[UUID],
-  flowId: UUID,
-  partitionId: UUID,
-  metadata: ControlMeasureMetadata,
-  runUniqueId: Option[String] = None, // todo why not UUID?
-  checkpoints: List[Checkpoint] = List.empty
-) extends BaseApiModel {
+  val flywayUrl = "jdbc:postgresql://localhost:5432/atum_db"
```
lsulak, line 21: I'd probably split the host, port, and DB name into 3 separate vals and add them here via s-interpolation

server/src/test/scala/za/co/absa/atum/server/api/TestTransactorProvider.scala
```diff
@@ -24,7 +24,7 @@ import zio.interop.catz._
 
 object TestTransactorProvider {
 
-  val layer: ZLayer[Any, Config.Error, Transactor[Task]] = ZLayer {
+  val layerWithoutRollback: ZLayer[Any, Config.Error, Transactor[Task]] = ZLayer {
```
lsulak, line 27: Todo q for asking - how these layers are gonna be loaded

server/src/main/scala/za/co/absa/atum/server/api/database/DoobieImplicits.scala
salamonpavel: Discussed on the call, I think we can resolve it.

project/Dependencies.scala
salamonpavel: thanks, extracted the version to Versions object

project/FlywayConfiguration.scala
salamonpavel: done

lsulak (dismissed):

lsulak (commented):
Please update the main Readme in the repo as well, it still contains a Spring section

server/README.md
```diff
@@ -1,23 +1,20 @@
 # Atum Web Service
 
-## How to build and run
+## How to build and run (will be updated soon)
```
lsulak, line 3: hmm, what do you mean will be updated soon?

salamonpavel: I mean the part about packaging for deployment. The first code block.

lsulak (approved):
Good to go in my view!

server/src/main/scala/za/co/absa/atum/server/api/controller/CheckpointController.scala
```diff
@@ -0,0 +1,53 @@
...
+import za.co.absa.atum.server.model.{ErrorResponse, GeneralErrorResponse, InternalServerErrorResponse}
+import zio._
+import zio.macros.accessible
+
+@accessible
+trait CheckpointController {
+  def createCheckpoint(checkpointDTO: CheckpointDTO): IO[ErrorResponse, CheckpointDTO]
+}
+
+class CheckpointControllerImpl(checkpointService: CheckpointService) extends CheckpointController {
```
yruslan, line 31: Is it a usual ZIO pattern to have both the trait and the class in the same file?

yruslan (commented):
Hey, I didn't make in time before the merge 😄 

This is very cool and very educating form me!

Just have one question of curiosity.

server/src/main/scala/za/co/absa/atum/server/api/controller/CheckpointController.scala
salamonpavel: @yruslan 
I haven't encountered any specific guidelines related to ZIO for this particular matter. Personally, I lean towards keeping interfaces or traits in separate files. However, I've noticed that this practice isn't consistently followed in the Atum project. Perhaps having these interfaces in the same files might better illustrate how to construct ZIO services. We can definitely reconsider this approach in the future, especially if we bring in other implementations. I'm curious to hear your thoughts on this topic. What's your perspective?

benedeki: This is actually funny. I saw this approach first and Hector, who did it in Atum Agent.
I am personally not that big of a fan either, but didn't consider it serious enough at that stage of development. Then I saw you @salamonpavel  using it and thought, it's one of the conventions out there. So I didn't bring it up anymore.

if we agree we prefer the style of having them in separate files, as it seems, let's do it 😄

yruslan: I'm on the same page. Having them separately is probably just a habit for me over some time. Definitely not a big deal in Scala, and if it better illustrates how ZIO works it is a good reason. For me class name having the same name as train/class inside is more for 2 reasons:
- It also going to match .class files in target
- IntelliJ idea does not place good icon on files for which class and file name do not match 😆  :
  
Screenshot 2024-02-09 at 12 44 54

So overall no strong opinions, especially for Atum, completely up to you.

PR #153, opened 2024-02-02
aws secrets provider

### LINKED ISSUE #107: Server -> DB: Add pulling DB login credentials via AWS Secrets

salamonpavel:
Reads credentials from AWS Secrets Manager. 
Closes #107

### DIFF
--- a/server/src/main/resources/reference.conf
+++ b/server/src/main/resources/reference.conf
@@ -12,7 +12,7 @@
   }
   aws {
     region = "af-south-1"
-    serviceUserSecretKey = "serviceUserSecretKey"
+    dbPasswordSecretName = "serviceUserSecretKey"
   }
   ssl {
     enabled=false
--- a/server/src/main/resources/reference.conf
+++ b/server/src/main/resources/reference.conf
@@ -1,4 +1,8 @@
 {
+   aws {
+     region = "af-south-1"
+     serviceUserSecretKey = "serviceUserSecretKey"
+   }
    logger {
       # This format includes timestamp, level, thread (fiberId), message, and cause
       format = "%label{timestamp}{%fixed{32}{%timestamp}} %label{level}{%level} %label{thread}{%fiberId} %label{class}{%name:%line} %label{message}{%message} %label{cause}{%cause}"
--- a/server/src/main/scala/za/co/absa/atum/server/aws/AwsSecretsProvider.scala
+++ b/server/src/main/scala/za/co/absa/atum/server/aws/AwsSecretsProvider.scala
@@ -1,46 +1,9 @@
-/*
- * Copyright 2021 ABSA Group Limited
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


server/src/main/scala/za/co/absa/atum/server/aws/AwsSecretsProvider.scala
```diff
@@ -0,0 +1,46 @@
...
+      secretsManagerClient.getSecretValue(getSecretValueRequest).secretString()
+    }
+  }
+}
+
+object AwsSecretsProviderImpl {
+  val layer: ZLayer[Any, Config.Error, AwsSecretsProviderImpl] = ZLayer {
+    for {
+      awsConfig <- ZIO.config[AwsConfig](AwsConfig.config)
+    } yield new AwsSecretsProviderImpl(SecretsManagerClient.builder().region(Region.of(awsConfig.region)).build())
```
TebaleloS, line 44: I like the fact that you used the region from the config file, very awesome!!!

server/src/main/resources/reference.conf
```diff
@@ -10,6 +10,10 @@
     # maximum number of connections that HikariCP will keep in the pool, including both idle and in-use connections
     maxPoolSize=10
   }
+  aws {
+    region = "af-south-1"
+    serviceUserSecretKey = "serviceUserSecretKey"
```
lsulak, line 15: I would definitely mention DB here. This is for secret for accessing database

```suggestion
    awsSecretNameForDBServiceUserPwd = "changeme"
```

or something like that

lsulak: I was actually not entirely sure I like this to be here, as I think that it's closer to the DB. But I understand that we might have more AWS stuff and it can belong close to each other for having better visibility of what's used from AWS...but hmm on the other hand, if working with the DB, such config-using code would be using other unrelated things, such as config for SSM or so...

I guess I'm fine with this currently

lsulak (commented):
Apart from the config name for the DB's PWD in AWS Secrets Manager, it looks really good and fully functional.

- [x]  code reviewed
- [x]  pulled
- [x]  built
- [x]  ran tests
- [x] ran on localhost with real configuration, created partitioning against local server and checked DEV DB - it all worked very nicely end to end

server/src/main/resources/reference.conf
salamonpavel: You are right, it wasn't obvious it's related to a database. And as you mentioned, it's difficult to foresee how things will evolve, what sort of AWS services we will be using. And even then we will face the dilemma of whether to group it by service or by technology. Anyway I have change the parameter to **dbPasswordSecretName** as I think it makes it clear and still isn't crazy long. Hope that's acceptable.

lsulak (approved):
Good to go!

server/src/main/resources/reference.conf
lsulak: it is, thanks it looks  better :)

PR #157, opened 2024-02-13
defined and implemented get_partitioning_measures:#137

### LINKED ISSUE #137: Create DB function to gather partitioning measure definitions
## Background
We need to read the partitioning information.

## Feature
Create DB function to gather partitioning measure definitions.

## Proposed Solution
Function signature
```PLpgSQL
CREATE OR REPLACE FUNCTION runs.get_partitioning_measures(
    IN  i_partitioning                      JSONB,
    OUT measure_name                        TEXT,
    OUT measured_columns                    TEXT[]
) RETURNS SETOF record AS

```

TebaleloS:
* Defined and implemented `get_partitioning_measures`.
* Added test cases but feel free to test with any size of the partitioning, as long as that partitioning exist in the database.

closes #137


database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,63 @@
...
+-------------------------------------------------------------------------------
+
+DECLARE
+    _fk_partitioning                    BIGINT;
+BEGIN
+    _fk_partitioning = runs._get_id_partitioning(i_partitioning);
+
+    IF _fk_partitioning IS NULL THEN
+        measure_name := 'Partitioning not found';
+        measured_columns := '{}';
```
lsulak, line 46: don't forget to drop these 2 since you are passing this info to the status and status text

TebaleloS: But that will complain that the `structure of query does not match function result type`.

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
```diff
@@ -0,0 +1,50 @@
...
+    function(fncGetPartitioningMeasures)
+      .setParam("i_partitioning", partitioning)
+      .execute { queryResult =>
+        val results = queryResult.distinct.next()
+        assert(results.getInt("status").contains(11))
+        assert(results.getString("status_text").contains("OK"))
+      }
+  }
+
+}
```
lsulak, line 50: Good work! Now, let's try to test more cases that the given DB function can be in:
* partitioning doesn't exist in the DB yet
* partitioning exists but there are no measures defined on it yet
* partitioning exists and there are measures on it - check that all measures were successfully returned

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
```diff
@@ -0,0 +1,50 @@
...
+      |""".stripMargin
+  )
+
+  test("Get partitioning measures") {
+    function(fncGetPartitioningMeasures)
+      .setParam("i_partitioning", partitioning)
+      .execute { queryResult =>
+        val results = queryResult.distinct.next()
+        assert(results.getInt("status").contains(11))
+        assert(results.getString("status_text").contains("OK"))
```
lsulak, line 46: I would also check which measures were returned. But for this you have to import them into the DB. Also you have to import partitioning into the DB, which is not done yet.

Balta has a very nice interface for this, please see WriteCheckpointTest file, and function calls like `table` . `import` or something like that

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
lsulak: Hmm, let's start with `measure_name := '';` then at least. I'll take a look whether it's possible to have some default nulls or so

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,63 @@
...
+        status := 41;
+        status_text := 'The partitioning does not exist.';
+        RETURN;
+    END IF;
+
+    status = 11;
```
lsulak, line 52: I would consider implementing another (third) status code, for situation when there are no measures in DB yet

lsulak: Based on this it will be something in range 40-49.....perhaps 42: https://github.com/AbsaOSS/fa-db/blob/master/core/src/main/scala/za/co/absa/fadb/status/README.md

What do you think @benedeki ?

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,63 @@
...
+-- Function: runs.get_partitioning_measures(2)
+--      Iterates over a JSONB object and returns each key-value pair as a record
+--
+-- Parameters:
+--      i_partitioning      - JSONB object where each key is a measure name and its corresponding value is an array of measured columns
+--
+-- Returns:
+--      measure_name        - Name of the measure
+--      measured_columns    - Array of columns associated with the measure
+--
```
benedeki, line 36: Missing info on returned statuses.

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,63 @@
...
+    _fk_partitioning                    BIGINT;
+BEGIN
+    _fk_partitioning = runs._get_id_partitioning(i_partitioning);
+
+    IF _fk_partitioning IS NULL THEN
+        measure_name := 'Partitioning not found';
+        measured_columns := '{}';
+        status := 41;
+        status_text := 'The partitioning does not exist.';
+        RETURN;
```
benedeki, line 49: If a Pl/PgSql function is returning a `SETOF` there are two ways how to return rows from the function.
1. Using `RETURN NEXT;` The current values of the output variable create one output row.
2. Using `RETURN QUERY` followed by a standard SQL query, most usually `SELECT ....`
In this case here I would use `RETURN NEXT;`

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
benedeki: The third status code might be useful for handling in fa-db Or actually might be complicating it. I am not sure in the current state of fa-db lib 😊 
But if a third status code is introduced, it should definitely be of the OK status range 10-19 ( I would suggest 16.). Because it's totally OK not to have any measures defined.

database/src/test/scala/za/co/absa/atum/database/runs/CreatePartitioningIfNotExists.scala
```diff
@@ -64,7 +64,7 @@ class CreatePartitioningIfNotExists extends DBTestSuite{
         row.getLong("id_partitioning").get
       }
 
-    table("runs.partitionings").where(add("id_partitioning", partitioningID)) {partitioningResult =>
+    table("runs.partitionings").where(add("id_partitioning", partitioningID)) { partitioningResult =>
```
benedeki, line 67: Probably an accidental change?

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
benedeki, line 46: If status is signaling an error, per our convention the content of the fields other then `status` and `status_text` is undefined (can be anything). In this case I would keep them as `NULL` (no assignment needed).

benedeki: Let the variables without assignment. They will be returned as `NULL`.

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
TebaleloS: Yah, it makes sense that we can have a partitioning without measures associated with it.

database/src/test/scala/za/co/absa/atum/database/runs/CreatePartitioningIfNotExists.scala
TebaleloS: Yes. I will roll back the changes.

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -33,29 +33,48 @@ $$
...
-        RETURN;
+        RETURN NEXT;
     END IF;
 
-    status = 11;
-    status_text = 'OK';
-
-    RETURN QUERY
-    SELECT md.measure_name, md.measured_columns, status, status_text
+    SELECT md.measure_name, md.measured_columns
```
benedeki, line 61: The original was correct. This way you would return only the first measure.
(The other option would be to cycle over the result of the `SELECT` query and do a `RETURN NEXT` for each row. Very ineffective.)

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -33,29 +33,48 @@ $$
 -- Returns:
 --      measure_name        - Name of the measure
 --      measured_columns    - Array of columns associated with the measure
+--      status              - Status code
+--      status_text         - Status message
+
+-- Status codes:
+--      10 - Record not found for the given partitioning
```
benedeki, line 40: **Minor:**
I would make this 16, to avoid the numbers we have [convention for](https://github.com/AbsaOSS/fa-db/blob/master/core/src/main/scala/za/co/absa/fadb/status/README.md).

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -33,29 +33,48 @@ $$
...
+    IF NOT FOUND THEN
+        measure_name := NULL;
+        measured_columns := NULL;
+        status := 10;
+        status_text := 'No measures found for the given partitioning.';
+    ELSE
+        measure_name := _measure_name;
+        measured_columns := _measured_columns;
+        status = 11;
+        status_text = 'OK';
```
benedeki, line 75: Not needed, if using `RETURN QUERY` above as suggested..
Only call `RETURN NEXT;` once to return one row signaling no measures exist.
```suggestion
    ELSE
        RETURN NEXT;
```

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -33,29 +33,48 @@ $$
...
+    END IF;
+    RETURN NEXT;
```
benedeki, line 77: And this should not be here then...

Customarily there's a RETURN statement at the end. But not necessary, as it's done automatically anyway.

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
```diff
@@ -18,33 +18,112 @@ package za.co.absa.atum.database.runs
...
+    function(fncGetPartitioningMeasures)
+      .setParam("i_partitioning", partitioning)
+      .execute { queryResult =>
+        val results = queryResult.distinct.next()
+        assert(results.getInt("status").contains(41))
+        assert(results.getString("status_text").contains("The partitioning does not exist."))
+      }
+  }
+
+  test("Get partitioning measures should return exception for partitioning with no measures") {
```
benedeki, line 93: Wrong name of the test, it's not an error.

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
benedeki: Still missing test where it returns some measures. It would have disclosed the mistake described above.

benedeki (changes requested):

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,82 @@
...
+     IN i_partitioning          JSONB,
+     OUT measure_name           TEXT,
+     OUT measured_columns       TEXT[],
+     OUT status                 INTEGER,
+     OUT status_text            TEXT
+ ) RETURNS SETOF record AS
+$$
+-------------------------------------------------------------------------------
+--
+-- Function: runs.get_partitioning_measures(2)
```
lsulak, line 27:
```suggestion
-- Function: runs.get_partitioning_measures(1)
```

this is number of input params

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,82 @@
...
+--      Iterates over a JSONB object and returns each key-value pair as a record
+--
+-- Parameters:
+--      i_partitioning      - JSONB object where each key is a measure name and its corresponding value is an array of measured columns
```
lsulak, line 31: not true

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,82 @@
...
+DECLARE
+    _fk_partitioning                    BIGINT;
+    _measure_name                       TEXT;
+    _measured_columns                   TEXT[];
+BEGIN
+    _fk_partitioning = runs._get_id_partitioning(i_partitioning);
+
+    IF _fk_partitioning IS NULL THEN
+        measure_name := NULL;
+        measured_columns := NULL;
```
lsulak, line 55: if you delete these 2 lines, I think that the default is gonna be NULL so you don't have to explicitly assign it. In Declarative world (languages such as SQL) = if a value doesn't exist / is not defined, then a NULL is returned, instead of an exception as it would be in Imperative languages

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,82 @@
...
+        status := 41;
+        status_text := 'The partitioning does not exist.';
+        RETURN NEXT;
```
lsulak, line 58: I would just cut it here and wouldn't continue - just return;

```suggestion
        RETURN;
```

lsulak: This returned when I tested it against non-existing partitioning. The second message is obvious so not needed in my opinion

Screenshot 2024-02-23 at 18 30 43

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,82 @@
...
+    END IF;
+
+    SELECT md.measure_name, md.measured_columns
+    INTO _measure_name, _measured_columns
+    FROM runs.measure_definitions AS md
+    WHERE md.fk_partitioning = _fk_partitioning;
+
+    IF NOT FOUND THEN
+        measure_name := NULL;
+        measured_columns := NULL;
```
lsulak, line 68: here as well I think you can drop these 2

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,82 @@
...
+        status := 10;
+        status_text := 'No measures found for the given partitioning.';
+    ELSE
+        measure_name := _measure_name;
+        measured_columns := _measured_columns;
+        status = 11;
+        status_text = 'OK';
+    END IF;
+    RETURN NEXT;
```
lsulak, line 77: why `RETURN NEXT` here?

database/src/main/postgres/runs/V1.5.11__get_partitioning_measures.sql
```diff
@@ -0,0 +1,82 @@
...
+END;
+$$
+LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
+
+ALTER FUNCTION runs.get_partitioning_measures(JSONB) OWNER TO atum_owner;
```
lsulak, line 82: I think you are missing this here:

`GRANT EXECUTE ON FUNCTION runs.get_partitioning_measures(JSONB) TO atum_user;`

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
```diff
@@ -0,0 +1,129 @@
...
+        .add("created_by", "Thomas")
+        .add("measure_name", s"measure${fkPartitioning}")
+        .add("measured_columns", CustomDBType("""{"col1"}""", "TEXT[]"))
+    )
+
+    function(fncGetPartitioningMeasures)
+      .setParam("i_partitioning", partitioning)
+      .execute { queryResult =>
+        val results = queryResult.distinct.next()
+        assert(results.getInt("status").contains(11))
```
lsulak, line 60: please also check whether the desired values in `measure_name` and `measured_columns` are there

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  test("Get partitioning measures should return error on non existing partitioning") {
+    val partitioning = JsonBString(
+      """
+        |{
+        |   "version": 1,
+        |   "keys": ["key1"],
+        |   "values": {
```
lsulak, line 77:
```suggestion
        |   "keysToValues": {
```

lsulak: partitioning with `values` is no longer valid. Did you see my changes in https://github.com/AbsaOSS/atum-service/pull/109 ❓

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
```diff
@@ -0,0 +1,129 @@
...
+        |}
+        |""".stripMargin
+    )
+
+    function(fncGetPartitioningMeasures)
+      .setParam("i_partitioning", partitioning)
+      .execute { queryResult =>
+        val results = queryResult.distinct.next()
+        assert(results.getInt("status").contains(41))
+        assert(results.getString("status_text").contains("The partitioning does not exist."))
```
lsulak, line 89: also you can add `assert(queryResult.hasNext)` in the end

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
```diff
@@ -0,0 +1,129 @@
...
+    )
+
+    val fkPartitioning: Long = table("runs.partitionings").fieldValue("partitioning", partitioning, "id_partitioning").get.get
+
+    function(fncGetPartitioningMeasures)
+      .setParam("i_partitioning", partitioning)
+      .execute { queryResult =>
+        val results = queryResult.distinct.next()
+        assert(results.getInt("status").contains(10))
+        assert(results.getString("status_text").contains("No measures found for the given partitioning."))
```
lsulak, line 124: you can add also these there:

`assert(!queryResult.hasNext)`

and check whether these are NULLs: `measure_name` and `measured_columns`

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningMeasuresTest.scala
TebaleloS: Done

benedeki:
Resolving addressed issues in pair programming session.

benedeki (approved):
Clarified things together. Looks good now, tests passed locally.

PR #161, opened 2024-02-20
endpoint for health checks

salamonpavel:
Added endpoint for health checks at path /health.

### DIFF
--- a/server/src/main/scala/za/co/absa/atum/server/Constants.scala
+++ b/server/src/main/scala/za/co/absa/atum/server/Constants.scala
@@ -17,11 +17,18 @@
 package za.co.absa.atum.server
 
 object Constants {
-  final val Api = "api"
-  final val V1 = "v1"
-  final val CreatePartitioning = "createPartitioning"
-  final val CreateCheckpoint = "createCheckpoint"
-  final val Health = "health"
+
+  object Endpoints {
+
+    final val Api = "api"
+    final val V1 = "v1"
+    final val CreatePartitioning = "createPartitioning"
+    final val CreateCheckpoint = "createCheckpoint"
+    final val Health = "health"
+
+  }
+
   final val SwaggerApiName = "Atum API"
   final val SwaggerApiVersion = "1.0"
+
 }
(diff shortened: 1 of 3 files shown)


server/src/main/scala/za/co/absa/atum/server/Constants.scala
```diff
@@ -21,6 +21,7 @@ object Constants {
   final val V1 = "v1"
   final val CreatePartitioning = "createPartitioning"
   final val CreateCheckpoint = "createCheckpoint"
+  final val Health = "health"
```
lsulak, line 24: Maybe I would consider this rather :D 

```suggestion
  final val HealthEndpoint = "health"
```

or to have a nested object called Endpoints

salamonpavel: I don't know. I welcome the initiative to make it more structured somehow. But maybe it's not quite the time for it yet. I want to avoid calling it HealthEndpoint as it's not an endpoint actually, just it's name and/or url path.

```scala
object Constants {
  final val Api = "api"
  final val V1 = "v1"
  final val CreatePartitioning = "createPartitioning"
  final val CreateCheckpoint = "createCheckpoint"
  final val Health = "health"
  final val SwaggerApiName = "Atum API"
  final val SwaggerApiVersion = "1.0"
}

object Constants {

  object Endpoints {

    final val Api = "api"
    final val V1 = "v1"
    final val CreatePartitioning = "createPartitioning"
    final val CreateCheckpoint = "createCheckpoint"
    final val Health = "health"

  }

  final val SwaggerApiName = "Atum API"
  final val SwaggerApiVersion = "1.0"

}

object Constants {
  
  object Endpoints {
    
    final val Api = "api"
    final val V1 = "v1"
    final val CreatePartitioning = "createPartitioning"
    final val CreateCheckpoint = "createCheckpoint"
    final val Health = "health"
    
  }
  
  object Swagger {

    final val SwaggerApiName = "Atum API"
    final val SwaggerApiVersion = "1.0"
    
  }
  
}
```

Perhaps we can revisit this at some future point?

lsulak: I quite like the second proposal, I think that the Endpoints can already be split

lsulak (approved):

PR #165, opened 2024-03-07
Custom (Uknown) Measure 

### LINKED ISSUE #99: Add a possibility to let the user specify a custom Measure

salamonpavel:
Introduces `UknownMeasure` type for data provided by client applications in case such a measure doesn't exist in Atum yet.
Closes #99


benedeki (commented):
Wanted to approve, but it fails on my local build...
Can somebody else try, please

PR #168, opened 2024-03-11
Feature/improve testability of agent

filiphornak:


agent/src/main/scala/za/co/absa/atum/agent/dispatcher/Capture.scala
```diff
@@ -0,0 +1,92 @@
+package za.co.absa.atum.agent.dispatcher
+
+import za.co.absa.atum.model.dto._
+
+import java.util
+import scala.jdk.CollectionConverters.IteratorHasAsScala
+
+class Capture(maxEvents: Int) extends Dispatcher {
```
miroslavpojer, line 8: I do not see usage of this class or object. What is the goal of it?

benedeki:
Done in #182

PR #167, opened 2024-03-11
Get Atum Context, by getting existing partitioning from DB - retrieve Measures and AdditionalData for a given partitioning #120

TebaleloS:
# Partitioning DB-Functions Calls

* Added class to call partitioning measures function
* Added class to call partitioning additional data DB function
* Defined two methods `getPartitioningMeasures` and `getPartitioningAdditionalData` and implemented them.
* Defined and implemented a method in the service `returnAtumContext` to package the above two methods including `createPartitioningIfNotExists` method in order to return the desired atum context.
* Added few test-cases for the newly added code.
* Included some test cases for `createOrUpdateAdditionalData` method as well.

closes #120

PR #192, opened 2024-04-22
#187: get all checkpoints for a flow

### LINKED ISSUE #187: Create DB function to get all checkpoint data of a flow
## Background
We need to read all the measurements across a checkpoint of certain name within a flow.

## Feature
Create a DB function that will return all checkpoint and its measurements for the given flow (identified by main partitioning)

## Proposed Solution
Function signature
```plpgsql
CREATE OR REPLACE FUNCTION flows.get_flow_checkpoints(
    IN i_flow_partitioning  JSONB,
    IN i_limit             INT DEFAULT 5,
    IN i_checkpoint_name   TEXT DEFAULT NULL,
    OUT status             INTEGER,
    OUT status_text        TEXT,
    OUT id_checkpoint      UUID,
    OUT checkpoint_name    TEXT,
    OUT measure_name       TEXT,
    OUT measure_columns    TEXT[],
    OUT measurement_value  JSONB,
    OUT checkpoint_time    TIMESTAMP WITH TIME ZONE
) RETURNS SETOF record AS

```

for output order by checkpoint_time  and id_checkpoint 

`i_limit` and i_checkpoint_name` are optional. If `NULL` no limit and no filter on checkpoint name.

Depends on #178

lsulak:
What was done:
* added DB function that retrieves all checkpoints and their measurements related to the flow, identify-able by a partitioning
* added ITs for the above
* fixing some problems in our current migration scripts, i.e. flyWay was failing 

Closes #187


database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
```diff
@@ -0,0 +1,127 @@
...
+--      status_text         - Status text
+--      id_checkpoint       - id of retrieved checkpoint
+--      checkpoint_name     - name of retrieved checkpoint
+--      measure_name        - measure name associated with a given checkpoint
+--      measure_columns     - measure columns associated with a given checkpoint
+--      measurement_value   - measurement details associated with a given checkpoint
+--      checkpoint_time     - time
+--
+-- Status codes:
+--      11                  - Flow created
```
benedeki, line 66:
```suggestion
--      11                  - OK
```

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
```diff
@@ -0,0 +1,127 @@
...
+--      16                  - No checkpoints found
```
benedeki, line 67: Would not use that status code imho. No checkpoints -> empty result set.

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
```diff
@@ -0,0 +1,127 @@
...
+      AND CASE
+            WHEN i_checkpoint_name IS NOT NULL THEN CP.checkpoint_name = i_checkpoint_name
+            ELSE TRUE
+          END
+    ORDER BY CP.process_start_time,
+             CP.id_checkpoint
+    LIMIT CASE
+            WHEN i_limit <= 0 THEN NULL -- NULL means no limit will be applied, same as LIMIT ALL
+            ELSE i_limit
+          END;
```
benedeki, line 114:
```suggestion
    LIMIT nullif(i_limit, 0)
```
Or you can make it a separate variable to make the query look shorter and easier to understand.

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
```diff
@@ -0,0 +1,127 @@
...
+            FROM flows.flows AS F
+            JOIN flows.partitioning_to_flow AS PF
+              ON F.id_flow = PF.fk_flow
+            WHERE F.fk_primary_partitioning = _fk_partitioning
+              AND CP.fk_partitioning = PF.fk_partitioning
+          )
+      AND CASE
+            WHEN i_checkpoint_name IS NOT NULL THEN CP.checkpoint_name = i_checkpoint_name
+            ELSE TRUE
+          END
```
benedeki, line 108:
```suggestion
      AND (i_checkpoint IS NULL OR CP.checkpoint_name = i_checkpoint_name)
```

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
```diff
@@ -0,0 +1,127 @@
...
+    END IF;
+
+    RETURN QUERY
+    SELECT 11 AS status, 'OK' AS status_text,
+           CP.id_checkpoint, CP.checkpoint_name,
+           MD.measure_name, MD.measured_columns,
+           M.measurement_value,
+           CP.process_start_time AS checkpoint_start_time, CP.process_end_time AS checkpoint_end_time
+    FROM runs.checkpoints AS CP
+    JOIN runs.measurements AS M
```
benedeki, line 92: **Small:**
Would make it explicit it's an `INNER JOIN`

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
```diff
@@ -0,0 +1,127 @@
+/*
```
benedeki, line 1: Make the file name `V1.8.3_...` or `V1.9.1_...`
And dependent on #167

database/src/main/postgres/flows/V1.7.4__flows.alter.ddl
```diff
@@ -13,6 +13,6 @@
  * limitations under the License.
  */
 
-ALTER COLUMN fk_primary_partitioning SET NOT NULL;
+ALTER TABLE flows.flows ALTER COLUMN fk_primary_partitioning SET NOT NULL;
```
benedeki, line 16: Already fixed in #167

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
```diff
@@ -0,0 +1,127 @@
...
+
+    IF _fk_partitioning IS NULL THEN
+        status := 41;
+        status_text := 'The partitioning does not exist.';
+        RETURN NEXT;
+        RETURN;
+    END IF;
+
+    RETURN QUERY
+    SELECT 11 AS status, 'OK' AS status_text,
```
benedeki, line 86: Matter of taste, but this alignment is... 🙅 😄

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
lsulak: hmm these status texts in the code docs are not always 1:1 to real messages. So I just wanted to make it more human readable / understandable.

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
lsulak: There is (as usual) a reasoning behind this:
* consistency - get_partitioning_additional_data and get_partitioning_measures for example have them
* it doesn't hurt to have the end user more informed, it's just a few bytes on the network traffic

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
lsulak: every join (unless otherwise specified, e.g. `LEFT` or `RIGHT`) is `INNER` - I don't really prefer such 'waste' of space, everybody knows this in SQL world and thus in this case I don't like to be explicit :) 

(really, based on my experience almost no one these days writes INNER)

Do you insist on it though?

database/src/main/postgres/flows/V1.7.4__flows.alter.ddl
lsulak: I know, but I would still like to keep it here because of easy testing - without this, my PR as a whole won't 'pass' and flyway will crash.

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
lsulak: what would be better? :D

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
lsulak: damned, this really hurts :D

lsulak: I chose 1.9.1 so that it's less schysophrenic

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
benedeki: Wait what?

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
benedeki: It's being removed actually in #167  😉

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
benedeki: If you want the negative then I would create a `_real_limit` local variable filled before the actual query based on the value of `i_limit`.

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
benedeki: Don't insist. But our bubbles have not intersection it seems. All people I remember wrote SQL write `INNER JOIN`. 😄 
Even in the code I saw yesterday in Absa... (will send in priv)

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
benedeki: Didn't we talk about it? 😉 
```sql
SELECT *
FROM my_table
    INNER JOIN your_table
WHERE true
```
or if you prefer
```sql
SELECT *
FROM   my table
    INNER JOIN your_table
WHERE  true
```

For the record, I don't want to force you to change it. Just mentioning this one is... hurting me 🤣

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
lsulak: but still, it's extremely cheap once we have https://github.com/AbsaOSS/fa-db/pull/120 (what was the point of implementing it then?)

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
lsulak: We talked and the outcome I gathered out of it was that we want left align style - in both ways, that `INNER JOIN` line is in my view a bit nasty, but indeed this is a personal style. (I'm used to https://www.sqlstyle.guide/ but left alignment is not a hill I want to die on :D )

lsulak: actually :( it doesn't look right (to me) when I change it like this: (a partial snippet)

```sql
    SELECT 11 AS status, 'OK' AS status_text,
           CP.id_checkpoint, CP.checkpoint_name,
           MD.measure_name, MD.measured_columns,
           M.measurement_value,
           CP.process_start_time AS checkpoint_start_time, CP.process_end_time AS checkpoint_end_time
    FROM   runs.checkpoints AS CP
    JOIN   runs.measurements AS M
      ON   CP.id_checkpoint = M.fk_checkpoint
    JOIN   runs.measure_definitions AS MD
      ON   M.fk_measure_definition = MD.id_measure_definition
    WHERE  EXISTS (
             -- Identification of a flow and all partitionings related to it.
             SELECT 1
             FROM   flows.flows AS F
             JOIN   flows.partitioning_to_flow AS PF
               ON   F.id_flow = PF.fk_flow
             WHERE  F.fk_primary_partitioning = _fk_partitioning
               AND  CP.fk_partitioning = PF.fk_partitioning
           )
      AND  (i_checkpoint_name IS NULL OR CP.checkpoint_name = i_checkpoint_name)
```

database/src/main/postgres/flows/V1.7.5__get_flow_checkpoints.sql
benedeki: I was thinking about it. Really do remove the status 16. It would complicate the Scala code. It would handle the case when there's an OK status, but still cannot (and shouldn't) create the output class (all `NULL`).

lsulak: Agreed

benedeki (approved):

PR #199, opened 2024-05-22
response envelope

### LINKED ISSUE #197: Envelope for return types
Controllers, Repositories and Services intact, though the below changes have to be introduced.

```scala
object ApiResponse {

  sealed trait ApiResponse[T]

  case class SingleApiResponse[T](data: T) extends ApiResponse[T]
  case class MultiApiResponse[T](data: Seq[T]) extends ApiResponse[T]

  implicit class SingleApiResponseEnhancement[T](val data: T) extends AnyVal {
    def toSingleApiResponse: SingleApiResponse[T] = {
      SingleApiResponse(data)
    }
  }

  implicit class MultiApiResponseEnhancement[T](val data: Seq[T]) extends AnyVal {
    def toMultiApiResponse: MultiApiResponse[T] = {
      MultiApiResponse(data)
    }
  }

}

// in the Endpoints trait
protected val createOrUpdateAdditionalDataEndpoint
    : PublicEndpoint[AdditionalDataSubmitDTO, ErrorResponse, SingleApiResponse[AdditionalDataSubmitDTO], Any] = {
    apiV1.post
      .in(CreateOrUpdateAdditionalData)
      .in(jsonBody[AdditionalDataSubmitDTO])
//      .out(jsonBody[AdditionalDataSubmitDTO])
      .out(jsonBody[SingleApiResponse[AdditionalDataSubmitDTO]])
  }

// in the Routes trait
private def createAllServerRoutes(httpMonitoringConfig: HttpMonitoringConfig): HttpRoutes[HttpEnv.F] = {
    val metricsInterceptorOption: Option[MetricsRequestInterceptor[HttpEnv.F]] = {
      if (httpMonitoringConfig.enabled) Some(HttpMetrics.prometheusMetrics.metricsInterceptor()) else None
    }
    val endpoints = List(
      createServerEndpoint(createCheckpointEndpoint, CheckpointController.createCheckpoint),
      createServerEndpoint(createPartitioningEndpoint, PartitioningController.createPartitioningIfNotExists),
      // has to be mapped here
      createServerEndpoint(createOrUpdateAdditionalDataEndpoint, PartitioningController.createOrUpdateAdditionalData _ andThen(_.map(_.toSingleApiResponse))),
      createServerEndpoint(healthEndpoint, (_: Unit) => ZIO.unit),
    )
    ZHttp4sServerInterpreter[HttpEnv.Env](http4sServerOptions(metricsInterceptorOption)).from(endpoints).toRoutes
  }

// and additional json serde implicits
implicit def readsSingleApiResponse[T: Reads]: Reads[SingleApiResponse[T]] = Json.reads[SingleApiResponse[T]]
implicit def writesSingleApiResponse[T: Writes]: Writes[SingleApiResponse[T]] = Json.writes[SingleApiResponse[T]]
```

salamonpavel:
Introduces envelopes for API responses.
Closes #197


server/src/main/scala/za/co/absa/atum/server/model/PlayJsonImplicits.scala
```diff
@@ -93,4 +94,7 @@ object PlayJsonImplicits {
   implicit val readsAtumContextDTO: Reads[AtumContextDTO] = Json.reads[AtumContextDTO]
   implicit val writesAtumContextDTO: Writes[AtumContextDTO] = Json.writes[AtumContextDTO]
 
+  implicit def readsSingleSuccessResponse[T: Reads]: Reads[SingleSuccessResponse[T]] = Json.reads[SingleSuccessResponse[T]]
+  implicit def writesSingleSuccessResponse[T: Writes]: Writes[SingleSuccessResponse[T]] = Json.writes[SingleSuccessResponse[T]]
```
lsulak, line 98: I think that there is almost nothing going on in these 2 functions. Thus, `Reads` and `Writes` can be possibly combined into 1, i.e. using `Format`. Possibly something like:

```scala
implicit def formatSingleSuccessResponse[T: Format]: Format[SingleSuccessResponse[T]] =
   Json.format[SingleSuccessResponse[T]]
```

I haven't tried it. What do you think? Perhaps we can consolidate some Reads+Writes into Format here in this module

server/src/test/scala/za/co/absa/atum/server/api/http/CreateCheckpointEndpointIntegrationTest.scala
```diff
@@ -18,26 +18,27 @@ package za.co.absa.atum.server.api.http
...
-import zio.test._
+import za.co.absa.atum.server.model.ErrorResponse.{GeneralErrorResponse, InternalServerErrorResponse}
+import za.co.absa.atum.server.model.PlayJsonImplicits._
+import za.co.absa.atum.server.model.SuccessResponse.SingleSuccessResponse
 import zio._
-import za.co.absa.atum.server.model.PlayJsonImplicits.{readsCheckpointDTO, writesCheckpointDTO}
 import zio.test.Assertion.equalTo
+import zio.test._
 
 object CreateCheckpointEndpointIntegrationTests extends ZIOSpecDefault with Endpoints with TestData {
```
lsulak, line 36: name of the file doesn't reflect the object name, `s` is missing from filename

server/src/main/scala/za/co/absa/atum/server/model/PlayJsonImplicits.scala
salamonpavel: yeah seems that it's ok, compiles without any issues ... changed it for the envelopes' types ... for the rest well I would say we could maybe finally replace play-json with circe ...

server/src/main/scala/za/co/absa/atum/server/model/SuccessResponse.scala
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.server.model
+
+object SuccessResponse {
```
lsulak, line 19: What about `ErrorResponse` ? I guess that since you haven't implemented yet, this would be a part of some future ticket?

Also, some metadata on the base envelope level (that even `SuccessResponse` and `ErrorResponse` would share), such as `requestID` and `requestURI` ? Maybe at least the `requestID` could be addressed here in this ticket? I.e. server-side generation of these

salamonpavel: `ErrorResponse` type already exists.

server/src/main/scala/za/co/absa/atum/server/model/PlayJsonImplicits.scala
lsulak: Okay, I'm fine with either doing that now (either of these two :D up to you)  - please can you check if there is a ticket for circe replacement and if not then create one?

server/src/main/scala/za/co/absa/atum/server/model/SuccessResponse.scala
lsulak: My apologies, I overlooked it - okay, but there is no hierarchy between them, i.e. no common ancestor that very likely we'll have when we start adding the metadata into the envelope

lsulak: so basically by adding it and `requestId` there, I think that it's a good step forward that we can easily extend later

lsulak: Update: okay, understood, it seems that the only thing `could` be the `requestID` - let's check with the team then, I feel that we could push it forward quite easily hopefully

server/src/main/scala/za/co/absa/atum/server/model/PlayJsonImplicits.scala
lsulak: awesome, thanks!

server/src/main/scala/za/co/absa/atum/server/api/http/Endpoints.scala
```diff
@@ -17,41 +17,43 @@
...
 
-  protected val createCheckpointEndpoint: PublicEndpoint[CheckpointDTO, ErrorResponse, CheckpointDTO, Any] = {
+  protected val createCheckpointEndpoint
+    : PublicEndpoint[CheckpointDTO, ErrorResponse, SingleSuccessResponse[CheckpointDTO], Any] = {
     apiV1.post
       .in(CreateCheckpoint)
       .in(jsonBody[CheckpointDTO])
       .out(statusCode(StatusCode.Created))
-      .out(jsonBody[CheckpointDTO])
+      .out(jsonBody[SingleSuccessResponse[CheckpointDTO]])
```
lsulak, line 38: actually I don't know if we want to do this - for potential compatibility reasons

lsulak: let's talk about it on the internal MS chat

TebaleloS (approved):
[1] Pulled
[1] Tested
[1] Compiled

As much as I have approved this, can we please wait for David's PR about fixing the `cross-build` to be merged

server/src/main/scala/za/co/absa/atum/server/api/controller/PartitioningControllerImpl.scala
```diff
@@ -16,39 +16,53 @@
...
+    atumContextDTOEffect
+  }
+
+  override def createPartitioningIfNotExistsV2(
+    partitioningSubmitDTO: PartitioningSubmitDTO
+  ): IO[ErrorResponse, SingleSuccessResponse[AtumContextDTO]] = {
+    mapToSingleSuccessResponse(createPartitioningIfNotExistsV1(partitioningSubmitDTO))
   }
 
   override def createOrUpdateAdditionalData(
```
lsulak, line 58: I'm wondering whether we should rename these to `v2` already. Because now it's:
* v1
* v2
* none (but they represent v2 basically)

What do you think?

salamonpavel: renamed to createOrUpdateAdditionalDataV2

lsulak:
So probably the only thing now we have to do (after this PR, I'd say) is to make the agent compatible with v2 APIs. And also using circe for SerDe fully/everywhere, instead of the current approach.

lsulak (dismissed):

lsulak (approved):

PR #237, opened 2024-08-03
219: Suggestion - Doobie multiple results with aggregated status

benedeki:
Just a suggestion, but it seems to work. Even added a unit test passing before the change and after it too.


TebaleloS (commented):
This is great!!!

TebaleloS:
I have implemented these changes in my previous PR inspired by this. So, I think is safe to **close this PR**. And thank you for this.

PR #241, opened 2024-08-08
get checkpoint

### LINKED ISSUE #231: implement functionality for getPartitioningCheckpointEndpointV2
```scala
protected val getPartitioningCheckpointEndpointV2
    : PublicEndpoint[(Long, String), ErrorResponse, SingleSuccessResponse[CheckpointDTO], Any] = {
    apiV2.get
      .in(V2Paths.Partitionings / path[Long]("partitioningId") / V2Paths.Checkpoints / path[String]("checkpointId"))
      .out(jsonBody[SingleSuccessResponse[CheckpointDTO]])
  }
```

salamonpavel:
Implements GET endpoint for checkpoint retrieval
Closes #231


salamonpavel:
Release notes:
- Introduces GET endpoint for single checkpoint retrieval

server/src/main/scala/za/co/absa/atum/server/api/repository/CheckpointRepositoryImpl.scala
```diff
@@ -37,13 +44,63 @@ class CheckpointRepositoryImpl(writeCheckpointFn: WriteCheckpoint, writeCheckpoi
...
+    )
+      .map(_.flatten)
+      .flatMap { checkpointItems =>
+        ZIO
+          .fromEither(checkpointItemsToCheckpointV2DTO(checkpointItems))
+          .mapError(error => GeneralDatabaseError(error.getMessage))
+      }
+  }
+
+  private def checkpointItemsToCheckpointV2DTO(
```
lsulak, line 61: Why not to move this functionality directly to `GetPartitioningCheckpointV2.scala` ?

server/src/main/scala/za/co/absa/atum/server/model/GetCheckpointV2Args.scala
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.server.model
+
+import java.util.UUID
+
+case class GetCheckpointV2Args(
```
lsulak, line 21: I'm not a fan of the placement of this one. What about moving it to `GetPartitioningCheckpointV2.scala` ? Similarly as here: https://github.com/absa-group/AUL/pull/2844/files#diff-e4b451049f88dfb52b116688e6b5e7e7baec175ce718a821d513d1b02aa75da2, just put it into the companion object of the function. I really like it - it's much more closer together and thus we can avoid having these random small case classes in service's internal model

server/src/main/scala/za/co/absa/atum/server/model/CheckpointItemFromDB.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package za.co.absa.atum.server.model
+
+import io.circe.Json
+
+import java.time.ZonedDateTime
+import java.util.UUID
+
+case class CheckpointItemFromDB(
```
lsulak, line 24: the same thing about the placement of these 'fromDB data containers' can be applied here as well :)

server/src/main/scala/za/co/absa/atum/server/api/http/Routes.scala
```diff
@@ -60,6 +62,16 @@ trait Routes extends Endpoints with ServerOptions {
         createOrUpdateAdditionalDataEndpointV2,
         PartitioningController.createOrUpdateAdditionalDataV2
       ),
+      createServerEndpoint[
+        (Long, UUID),
```
lsulak, line 66: sometimes types are here and sometimes not, why's that?

server/src/main/scala/za/co/absa/atum/server/api/http/Endpoints.scala
```diff
@@ -75,6 +77,15 @@ trait Endpoints extends BaseEndpoints {
       .out(jsonBody[SingleSuccessResponse[AdditionalDataSubmitDTO]])
   }
 
+  protected val getPartitioningCheckpointEndpointV2
+    : PublicEndpoint[(Long, UUID), ErrorResponse, SingleSuccessResponse[CheckpointV2DTO], Any] = {
```
lsulak, line 81: GetCheckpointV2Args can be used here

server/src/main/scala/za/co/absa/atum/server/api/database/runs/functions/WriteCheckpointV2.scala
```diff
@@ -31,18 +31,18 @@ import za.co.absa.db.fadb.doobie.postgres.circe.implicits.jsonbArrayPut
...
-    Seq(
-      fr"${args.partitioningId}",
-      fr"${args.checkpointV2DTO.id}",
-      fr"${args.checkpointV2DTO.name}",
-      fr"${args.checkpointV2DTO.processStartTime}",
-      fr"${args.checkpointV2DTO.processEndTime}",
-      fr"${args.checkpointV2DTO.measurements.toList.map(_.asJson)}",
-      fr"${args.checkpointV2DTO.measuredByAtumAgent}",
-      fr"${args.checkpointV2DTO.author}"
+    extends DoobieSingleResultFunctionWithStatus[WriteCheckpointV2Args, Unit, Task](args =>
```
lsulak, line 34: I like the way you split partitioningId and generally speaking the `WriteCheckpointV2Args` structure 👍

database/src/main/postgres/runs/V1.9.3__get_partitioning_checkpoint_v2.sql
```diff
@@ -0,0 +1,109 @@
...
+        RETURN;
+    END IF;
+END;
+$$
+
+    LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
+
+ALTER FUNCTION runs.get_partitioning_checkpoint_v2(BIGINT, UUID) OWNER TO atum_owner;
+
+GRANT EXECUTE ON FUNCTION runs.get_partitioning_checkpoint_v2(BIGINT, UUID) TO atum_owner;
```
lsulak, line 109: new newline

lsulak (commented):
My review is finished on this one, I found a few things but otherwise it's a very good piece of work :)

server/src/main/scala/za/co/absa/atum/server/api/http/Routes.scala
salamonpavel: it's because sometimes compiler needs that defined explicitly as it cannot infer everything ...

server/src/main/scala/za/co/absa/atum/server/api/http/Endpoints.scala
salamonpavel: I can't be done that way. It has to be a tuple in this form. Feel free to experiment ....

server/src/main/scala/za/co/absa/atum/server/api/repository/CheckpointRepositoryImpl.scala
salamonpavel: I disagree. It's not a concern of the database function. It's simply a representation on Scala level of what happens on the DB level. The SQL code returns a certain data structure which is not CheckpointV2DTO. That's why we have repository layer in our application, we can do these mappings there. But if you feel we should move this particular method some place else then a companion of the CheckpointV2DTO could be one option. But I am not a big fan of that since CheckpointV2DTO is part of model module and I don't see a reason why should we expose this to agent module which depends on the model module. I think these so called DTOs should be used only at the boundaries of our application (for http communication). In other words at repository layer we should return CheckpointV2 and this class could have a method in its companion for conversion to DTO representation of this type. Unfortunately we haven't started doing it this way from the beginning but I would be happy to introduce this sort of convention of not mixing these data models. Let's discuss with others.

server/src/main/scala/za/co/absa/atum/server/model/GetCheckpointV2Args.scala
salamonpavel: Again, let's discuss with others. In general I am not against placing these in those companions but I can also imagine cases where we use the same structure for several database functions and then it wouldn't be appropriate to import from one fa-db object to another one. Personally I would rather create a separate package for the args classes.

server/src/main/scala/za/co/absa/atum/server/model/CheckpointItemFromDB.scala
salamonpavel: As above I would rather argue for a separate package as again multiple fa-db object could be returning the same data structure.

database/src/main/postgres/runs/V1.9.3__get_partitioning_checkpoint_v2.sql
```diff
@@ -0,0 +1,109 @@
...
+            runs.checkpoints C
+                JOIN
+            runs.measurements M ON C.id_checkpoint = M.fk_checkpoint
+                JOIN
+            runs.measure_definitions MD ON M.fk_measure_definition = MD.id_measure_definition
+        WHERE
+            C.fk_partitioning = i_partitioning_id
+          AND
+            C.id_checkpoint = i_checkpoint_id
+        ;
```
TebaleloS, line 94: Please put this to be in line with line 93

database/src/main/postgres/runs/V1.9.3__get_partitioning_checkpoint_v2.sql
```diff
@@ -0,0 +1,109 @@
...
+    IF NOT FOUND THEN
+        status := 42;
+        status_text := 'Checkpoint not found';
+        RETURN NEXT;
+        RETURN;
+    END IF;
+END;
+$$
+
+    LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
```
TebaleloS, line 105: Please remove the leading spaces on this, it looks weird.

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningCheckpointV2IntegrationTests.scala
```diff
@@ -0,0 +1,163 @@
...
+
+import java.time.OffsetDateTime
+import java.util.UUID
+
+class GetPartitioningCheckpointV2IntegrationTests extends DBTestSuite{
+
+  private val fncGetPartitioningCheckpointV2 = "runs.get_partitioning_checkpoint_v2"
+
+  case class MeasuredDetails(
+                              measureName: String,
```
TebaleloS, line 31: please align this

TebaleloS (dismissed):

PR #251, opened 2024-08-21
#250 - Update of filename inspection workflow

### LINKED ISSUE #250: Update of filename inspection workflow
## Background
The current implementation works with the master version of filename-inspector tools.
It is not the correct solution when the tool has already been released.

## Feature
Update the used version of the tool.
Improve too broad wildcard definition in exclusions.

miroslavpojer:
- Used released version of filename-inspector.
- Defined row definition of excludes.

Closes #250

### DIFF
--- a/.github/workflows/test_filenames_check.yml
+++ b/.github/workflows/test_filenames_check.yml
@@ -31,11 +31,14 @@ jobs:
 
       - name: Filename Inspector
         id: scan-test-files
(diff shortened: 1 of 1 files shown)


miroslavpojer:
Release note:
- Used the released version of filename-inspector.
- Defined row definition of excludes.

Zejnilovic (approved):

PR #258, opened 2024-09-03
Feature/224 post partitioning

### LINKED ISSUE #224: POST /partitionings
## Feature
 Implement functionality for `POST /partitionings` V2 for creating a partitioning in Atum Server

Body of the POST contains:
* the partitioning to create
* optionally the id of the parent partitioning (it must exist)

Returns 201and URI for the created entity

## Proposed Solution
```scala
case class PartitioningWithIdDTO (
  id: Long,
  partitioning: PartitioningDTO,
  parentPartitioning: Option[PartitioningDTO],
  author: String
)

object PartitioningWithIdDTO {
  implicit val encoderPartitioningWithIdDTO: Encoder[PartitioningWithIdDTO] = deriveEncoder
  implicit val decoderPartitioningWithIdDTO: Decoder[PartitioningWithIdDTO] = deriveDecoder
}

// Endpoints
protected val postPartitioningEndpointV2: PublicEndpoint[
    PartitioningSubmitDTO,
    ErrorResponse,
    (SingleSuccessResponse[PartitioningWithIdDTO], String),
    Any
  ] = {
    apiV2.post
      .in(V2Paths.Partitionings)
      .in(jsonBody[PartitioningSubmitDTO])
      .out(statusCode(StatusCode.Created))
      .out(jsonBody[SingleSuccessResponse[PartitioningWithIdDTO]])
      .out(header[String]("Location"))
      .errorOutVariantPrepend(conflictErrorOneOfVariant)
  }

// Routes
createServerEndpoint(postPartitioningEndpointV2, PartitioningController.postPartitioningEndpointV2)

// in PartitioningControllerImpl
override def postPartitioningEndpointV2(
    partitioningSubmitDTO: PartitioningSubmitDTO
  ): IO[ErrorResponse, (SingleSuccessResponse[PartitioningWithIdDTO], String)]

// in PartitioningServiceImpl
override def postPartitioning(
    partitioningSubmitDTO: PartitioningSubmitDTO
  ): IO[ServiceError, Either[StatusException, PartitioningWithIdDTO]]

// in PartitioningRepositoryImpl
override def postPartitioning(
    partitioningSubmitDTO: PartitioningSubmitDTO
  ): IO[DatabaseError, Either[StatusException, PartitioningWithIdDTO]]
```

Plus SQL code and fa-db class....

salamonpavel:
Introduces POST endpoint for partitioning creation
Closes #224


salamonpavel:
Release notes

- Introduces POST endpoint for partitioning creation

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+    OUT id_partitioning         BIGINT
+) RETURNS record AS
+$$
+    -------------------------------------------------------------------------------
+--
+-- Function: runs.create_partitioning(3)
+--      Creates a partitioning entry
+--
+-- Parameters:
+--      i_partitioning          - partitioning which existence to check
```
lsulak, line 32:
```suggestion
--      i_partitioning          - partitioning to create or which existence to check
```

perhaps something like this

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+--      31                  - Partitioning already present
+--
+-------------------------------------------------------------------------------
+DECLARE
+    _fk_parent_partitioning BIGINT := NULL;
+    _create_partitioning    BOOLEAN;
+    _status                 BIGINT;
+BEGIN
+
+id_partitioning := runs._get_id_partitioning(i_partitioning, true);
```
lsulak, line 53: the alignment of the fnc body needs to be shifted by 1 tab or something to the right

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+    RETURNING partitionings.id_partitioning
+        INTO create_partitioning.id_partitioning;
+
+    PERFORM 1
+    FROM flows._create_flow(id_partitioning, i_by_user);
+
+    status := 11;
+    status_text := 'Partitioning created';
+ELSE
+    status := 31;
```
lsulak, line 76: what, is this the only difference, the status code for situation when partitioning is already present? In comparison to `runs.create_partitioning_if_not_exists` function. Otherwise I think this is 1:1 copy almost

lsulak: so I think we shouldn't have same implementation twice..possibly just call `runs.create_partitioning_if_not_exists` from this function and change the status code in select / case when statement, if this is really needed

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
salamonpavel: done

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
salamonpavel: There is a difference in that the new function does register parent partitioning only when the child partitioning does not exist. Anyway I think we should have a call maybe about the create partitioning operation in general ...

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+
+        IF _create_partitioning THEN
+            -- copying measure definitions to establish continuity
+            INSERT INTO runs.measure_definitions(fk_partitioning, measure_name, measured_columns, created_by, created_at)
+            SELECT id_partitioning, CMD.measure_name, CMD.measured_columns, CMD.created_by, CMD.created_at
+            FROM runs.measure_definitions CMD
+            WHERE CMD.fk_partitioning = _fk_parent_partitioning;
+
+            -- additional data are not copied, they are specific for particular partitioning
+        ELSIF (_status = 11) THEN
```
benedeki, line 95: **Semi-important:**
This will change _status -> 11, Partitioning created_ too to status -> 12
Do we want that? 
Or maybe three OK statuses, 11, 12 and 13 fore paritioning created and linked to parent? 🤔

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+--
+-- Status codes:
+--      11                  - Partitioning created
+--      12                  - Partitioning parent registered
+--      31                  - Partitioning already present
+--
+-------------------------------------------------------------------------------
+DECLARE
+    _fk_parent_partitioning BIGINT := NULL;
+    _create_partitioning    BOOLEAN;
```
benedeki, line 49: A couple of suggestions to make the code easier to follow, IMHO (will be marked with 👓 )
👓
```suggestion
    _partitioning_created   BOOLEAN;
```

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+        PERFORM 1
+        FROM flows._create_flow(id_partitioning, i_by_user);
+
+        status := 11;
+        status_text := 'Partitioning created';
+    ELSE
+        status := 31;
+        status_text := 'Partitioning already present';
+        RETURN;
+    END IF;
```
benedeki, line 79: 👓
Move this block instead of line 55, while chaning the condition to
```sql
id_partitioning IS NULL
```

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+        INSERT INTO runs.partitionings (partitioning, created_by)
+        VALUES (i_partitioning, i_by_user)
+        RETURNING partitionings.id_partitioning
+            INTO create_partitioning.id_partitioning;
+
+        PERFORM 1
+        FROM flows._create_flow(id_partitioning, i_by_user);
+
+        status := 11;
+        status_text := 'Partitioning created';
```
benedeki, line 74:
```suggestion
        status_text := 'Partitioning created';
        _partitioning_created := TRUE;
```

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+    ELSE
+        status := 31;
+        status_text := 'Partitioning already present';
```
benedeki, line 77:
```suggestion
        status_text := 'Partitioning already present';
                _partitioning_created := TRUE;

```

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+        RETURN;
```
benedeki, line 78: **Important:**
Do we want to exit here. What if the goal is to only connect the parent?
But if parent connection will be done via other function and `RETURN;` is justified here, then some other code can be simplified. Particularly the `_partitioning_created` is not needed.

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+    END IF;
+
+    IF i_parent_partitioning IS NOT NULL THEN
```
benedeki, line 81: 👓
This is joined with the same condition above.

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,107 @@
...
+
+        SELECT ATPF.status
+        FROM flows._add_to_parent_flows(_fk_parent_partitioning, id_partitioning, i_by_user) AS ATPF
+        INTO _status;
+
+        IF _create_partitioning THEN
```
benedeki, line 87: 👓
```suggestion
        IF _partitioning_created THEN
```

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
benedeki: Let's find a spot tomorrow. I think/hope it won't be long.

salamonpavel:
The code has been updated. Most of previous comments about the SQL part have become obsolete as the function has been refactored and simplified. @benedeki and @lsulak

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
benedeki: Irrelevant after the changes.

server/src/main/scala/za/co/absa/atum/server/api/controller/PartitioningControllerImpl.scala
```diff
@@ -49,12 +50,6 @@ class PartitioningControllerImpl(partitioningService: PartitioningService)
     atumContextDTOEffect
   }
 
-  override def createPartitioningIfNotExistsV2(
```
benedeki, line 52: If you remove this one, won't it remove the v1 endpoint (make it inoperable?

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

server/src/main/scala/za/co/absa/atum/server/api/controller/PartitioningControllerImpl.scala
benedeki: According to @salamonpavel 
_There is createPartitioningIfNotExistsV1 method for the V1 endpoint._

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,93 @@
...
+        IF NOT FOUND THEN
+            status := 41;
+            status_text := 'Parent partitioning not found';
+            RETURN;
+        END IF;
+    END IF;
+
+    INSERT INTO runs.partitionings (partitioning, created_by)
+    VALUES (i_partitioning, i_by_user)
+    RETURNING partitionings.id_partitioning INTO create_partitioning.id_partitioning;
```
lsulak, line 68: I didn't know that these were possible :)

database/src/main/postgres/runs/V1.9.7__create_partitioning.sql
```diff
@@ -0,0 +1,93 @@
...
+$$
+-------------------------------------------------------------------------------
+--
+-- Function: runs.create_partitioning(3)
+--      Creates a partitioning entry
+--
+-- Parameters:
+--      i_partitioning              - partitioning to create or which existence to check
+--      i_by_user                   - user behind the change
+--      i_parent_partitioning_id    - (optional) parent partitioning id
```
lsulak, line 34:
```suggestion
--      i_parent_partitioning_id    - (optional) parent partitioning id, must already exist if specified 
```

or something like that

PR #265, opened 2024-09-05
Feature/230 get partitioning checkpoints

### LINKED ISSUE #230: GET /partitionings/{partId}/checkpoints
## Feature
Implement `GET /partitionings/{partId}/checkpoints` V2 REST API endpoint on the server

Returns sequence of Checkpoints with PAGINATION.

## Proposed Solution
Note: it should be IMHO concern of CheckpointsController/Service/Repository, not as currently the PartitioningController/Service/Repository
Note: Remove the initial get checkpoints endpoint

```scala
protected val getPartitioningCheckpointsEndpointV2
    : PublicEndpoint[(Long, Option[String], Option[Int], Option[Long]), ErrorResponse, PaginatedResponse[
      CheckpointDTO
    ], Any] = {
    apiV2.get
      .in(V2Paths.Partitionings / path[Long]("partitioningId") / V2Paths.Checkpoints)
      .in(query[Option[String]]("checkpoint-name"))
      .in(query[Option[Int]]("limit"))
      .in(query[Option[Long]]("offset"))
      .out(jsonBody[PaginatedResponse[CheckpointDTO]])
  }
```

salamonpavel:
Introduces GET partitioning checkpoints endpoint with pagination
Closes #230


salamonpavel:
Release notes:

- Introduces GET partitioning checkpoints endpoint with pagination

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,53 +28,64 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
+-- Function: runs.get_partitioning_checkpoints(4)
+--      Retrieves checkpoints (measures and their measurement details) related to a
 --      given partitioning (and checkpoint name, if specified).
 --
 -- Parameters:
 --      i_partitioning          - partitioning of requested checkpoints
---      i_limit                 - (optional) maximum number of checkpoint's measurements to return
---                                if 0 specified, all data will be returned, i.e. no limit will be applied
+--      i_limit                 - (optional) maximum number of checkpoints to return
+--      i_offset                - (optional) offset of the first checkpoint to return
```
lsulak, line 44: I would consider mentioning in this code doc - maybe here for `i_offset` even - based on which the checkpoints are being returned (the sorting logic)

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,53 +28,64 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
+    IF NOT FOUND THEN
         status := 41;
         status_text := 'Partitioning not found';
         RETURN NEXT;
         RETURN;
     END IF;
 
     RETURN QUERY
+        WITH limited_checkpoints AS (
+            SELECT DISTINCT C.id_checkpoint,
```
lsulak, line 80: Why this `DISTINCT` is here?

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,53 +28,64 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
-    RETURNS SETOF record AS
-$$
-    -------------------------------------------------------------------------------
+RETURNS SETOF record AS
+-------------------------------------------------------------------------------
 --
--- Function: runs.get_partitioning_checkpoints(JSONB, INT, TEXT)
---      Retrieves all checkpoints (measures and their measurement details) related to a
+-- Function: runs.get_partitioning_checkpoints(4)
+--      Retrieves checkpoints (measures and their measurement details) related to a
```
lsulak, line 38: I would explicitly document here in the fnc desc that one record being returned does not equal to a checkpoint, i.e. something that puts emphasis on fact that `i_limit` = X doesn't mean that no more than `X` records will be returned - it can be actually multiples of that, those pagination-related parameters & mechanism is all about checkpoints, not # of data returned, which might confuse someone

lsulak (commented):
LGTM, I like it, I just found a few tiny improvement ideas

server/src/main/scala/za/co/absa/atum/server/api/database/runs/functions/GetPartitioningCheckpoints.scala
```diff
@@ -17,31 +17,31 @@
...
-        fr"${values.checkpointName}"
+        fr"${args.partitioningId}",
+        fr"${args.limit}",
+        fr"${args.offset}",
+        fr"${args.checkpointName}"
       )
     )
     with StandardStatusHandling
-    with ByFirstErrorStatusAggregator {
+    with ByFirstRowStatusAggregator {
```
lsulak, line 44: Why this change?

lsulak: An error can get propagated silently

server/src/main/scala/za/co/absa/atum/server/model/CheckpointItemFromDB.scala
```diff
@@ -71,4 +72,22 @@ object CheckpointItemFromDB {
     }
   }
 
+  def groupAndConvertItemsToCheckpointV2DTOs(
+                                        checkpointItems: Seq[CheckpointItemFromDB]
+                                      ): Either[DecodingFailure, Seq[CheckpointV2DTO]] = {
```
lsulak, line 77: consider reformatting

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
salamonpavel: will remove, it's truly not needed

server/src/main/scala/za/co/absa/atum/server/api/database/runs/functions/GetPartitioningCheckpoints.scala
salamonpavel: The error cannot go unnoticed because the same status is set for all rows and that's also why I decided to use the byFirstRow approach.

server/src/main/scala/za/co/absa/atum/server/model/CheckpointItemFromDB.scala
salamonpavel: done

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
salamonpavel: ok added an explanation that the limit is related to number of checkpoints, not to amount of returned rows and that usually there is more than one row per checkpoint

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
salamonpavel: ok added note about these params in the docs

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,53 +28,69 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
         status_text := 'Partitioning not found';
         RETURN NEXT;
         RETURN;
     END IF;
 
     RETURN QUERY
+        WITH limited_checkpoints AS (
+            SELECT C.id_checkpoint,
+                   C.process_start_time,
+                   ROW_NUMBER() OVER (ORDER BY C.process_start_time DESC, C.id_checkpoint) AS rn
```
benedeki, line 87: This is a really cool trick. 😎 
But I think it would be more efficient to have a pre-query
```sql
IF i_limit IS NOT NULL THEN
    SELECT count(1) > i_limit
    FROM runs.checkpoints C
    WHERE C.fk_partitioning = i_partitioning_id
        AND (i_checkpoint_name IS NULL OR C.checkpoint_name = i_checkpoint_name)
    ORDER BY C.process_start_time DESC, C.id_checkpoint
    LIMIT i_limit + 1 OFFSET i_offset
    INTO _has_more;
ELSE
    _has_more := false;
END IF;
```
And then use  the `_has_more` in the query. The query will be actually cached by the PG engine.

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -86,28 +102,28 @@ BEGIN
...
+                JOIN
             runs.measurements M ON C.id_checkpoint = M.fk_checkpoint
-        JOIN
+                JOIN
             runs.measure_definitions MD ON M.fk_measure_definition = MD.id_measure_definition
         WHERE
-            C.fk_partitioning = _fk_partitioning
-        AND
-            (i_checkpoint_name IS NULL OR C.checkpoint_name = i_checkpoint_name)
+            C.id_checkpoint IN (SELECT LC.id_checkpoint FROM limited_checkpoints as LC WHERE LC.rn <= i_limit)
```
benedeki, line 114: Why you don't use the `limited_checkpoints` instead of `runs.checkpoints`? (Of course `limited_checkpoint` would need more columns in its `SELECT`)

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,53 +28,69 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
+    PERFORM 1 FROM runs.partitionings WHERE id_partitioning = i_partitioning_id;
+    IF NOT FOUND THEN
         status := 41;
         status_text := 'Partitioning not found';
         RETURN NEXT;
         RETURN;
     END IF;
 
     RETURN QUERY
+        WITH limited_checkpoints AS (
```
benedeki, line 84: Great solution 👍

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,53 +28,69 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
---      Retrieves all checkpoints (measures and their measurement details) related to a
+-- Function: runs.get_partitioning_checkpoints(4)
+--      Retrieves checkpoints (measures and their measurement details) related to a
 --      given partitioning (and checkpoint name, if specified).
 --
 -- Parameters:
 --      i_partitioning          - partitioning of requested checkpoints
---      i_limit                 - (optional) maximum number of checkpoint's measurements to return
---                                if 0 specified, all data will be returned, i.e. no limit will be applied
+--      i_limit                 - (optional) maximum number of checkpoints to return
```
benedeki, line 43:
```suggestion
--      i_checkpoints_limit                 - (optional) maximum number of checkpoints to return
```

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -86,28 +102,28 @@ BEGIN
...
         WHERE
-            C.fk_partitioning = _fk_partitioning
-        AND
-            (i_checkpoint_name IS NULL OR C.checkpoint_name = i_checkpoint_name)
+            C.id_checkpoint IN (SELECT LC.id_checkpoint FROM limited_checkpoints as LC WHERE LC.rn <= i_limit)
         ORDER BY
             C.process_start_time,
-            C.id_checkpoint
-        LIMIT nullif(i_limit, 0);
+            C.id_checkpoint;
```
benedeki, line 117: **Maybe premature optimization:**
Would consider having two branches of `RETURN QUERY` based if `i_limit` is NULL or not. (Without `i_limit` `WITH` is not required.

benedeki (commented):
Checked the SQL only so far.

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
lsulak: I like it

server/src/main/scala/za/co/absa/atum/server/api/database/runs/functions/GetPartitioningCheckpoints.scala
lsulak: hmm okay, in this case it's true

lsulak:
LGTM! I'd be interested in David's solution(s) regarding SQL code as well, but apart from that I like it

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
salamonpavel: ok changed to i_checkpoints_limit

lsulak (dismissed):

lsulak (approved):

benedeki:
Any chance to increase the coverage %, @salamonpavel ?

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,86 +28,113 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
-            C.fk_partitioning = _fk_partitioning
-        AND
-            (i_checkpoint_name IS NULL OR C.checkpoint_name = i_checkpoint_name)
         ORDER BY
-            C.process_start_time,
-            C.id_checkpoint
-        LIMIT nullif(i_limit, 0);
+            LC.process_start_time, LC.id_checkpoint;
 
+    IF NOT FOUND THEN
```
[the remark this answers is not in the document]
benedeki: Actually, so far in queries returning detail records of a master one, of there were no details one, we ust returned the empty set. Would prefer we stay consistent, and if for some reason this is considered better, changer there too.
Tag @TebaleloS @lsulak

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,86 +28,113 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
         RETURN NEXT;
         RETURN;
     END IF;
 
+    IF i_checkpoints_limit IS NOT NULL THEN
+        SELECT count(*) > i_checkpoints_limit
+        FROM runs.checkpoints C
+        WHERE C.fk_partitioning = i_partitioning_id
+          AND (i_checkpoint_name IS NULL OR C.checkpoint_name = i_checkpoint_name)
+        LIMIT i_checkpoints_limit + 1 OFFSET i_offset
```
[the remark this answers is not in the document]
salamonpavel: let's keep the code as is then for now :)

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,86 +28,113 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
+            LC.process_start_time AS checkpoint_start_time,
+            LC.process_end_time AS checkpoint_end_time,
+            _has_more AS has_more
         FROM
-            runs.checkpoints C
-        JOIN
-            runs.measurements M ON C.id_checkpoint = M.fk_checkpoint
-        JOIN
+            limited_checkpoints LC
+                JOIN
```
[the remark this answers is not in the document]
salamonpavel: ok, will add the 'INNER' and do so going forward always

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,86 +28,113 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
             runs.measure_definitions MD ON M.fk_measure_definition = MD.id_measure_definition
-        WHERE
-            C.fk_partitioning = _fk_partitioning
-        AND
-            (i_checkpoint_name IS NULL OR C.checkpoint_name = i_checkpoint_name)
         ORDER BY
-            C.process_start_time,
-            C.id_checkpoint
-        LIMIT nullif(i_limit, 0);
+            LC.process_start_time, LC.id_checkpoint;
```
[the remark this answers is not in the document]
salamonpavel: ok switched the order

server/src/main/scala/za/co/absa/atum/server/api/http/Endpoints.scala
```diff
@@ -108,12 +108,17 @@ trait Endpoints extends BaseEndpoints {
...
   protected val getPartitioningCheckpointsEndpointV2
-    : PublicEndpoint[CheckpointQueryDTO, ErrorResponse, MultiSuccessResponse[CheckpointDTO], Any] = {
+  : PublicEndpoint[(Long, Option[Int], Option[Long], Option[String]), ErrorResponse, PaginatedResponse[
+    CheckpointV2DTO
+  ], Any] = {
     apiV2.get
-      .in(GetPartitioningCheckpoints)
-      .in(jsonBody[CheckpointQueryDTO])
+      .in(V2Paths.Partitionings / path[Long]("partitioningId") / V2Paths.Checkpoints)
+      .in(query[Option[Int]]("limit").default(Some(10)).validateOption(Validator.inRange(1, 1000)))
```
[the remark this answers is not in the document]
salamonpavel: I agree we could maybe have some base endpoint for all/most paginated endpoints so we don't have to repeat this for every single one of them. Didn't want to optimize it yet given it's the first paginated endpoint ...

salamonpavel:
> > Any chance to increase the coverage %, @salamonpavel ?
> 
> Coincidentally @miroslavpojer today mentioned that jacoco-report had release version 1.7.1 with number of bugifxes. Can you please increase the version in the script to test and see how the new version works.

Those coverage numbers are incorrect.

miroslavpojer:
> > > Any chance to increase the coverage %, @salamonpavel ?
> > 
> > 
> > Coincidentally @miroslavpojer today mentioned that jacoco-report had release version 1.7.1 with number of bugifxes. Can you please increase the version in the script to test and see how the new version works.
> 
> Those coverage numbers are incorrect.

The new jacoco-report version explains the meaning of the negative percentage numbers.

`The "delta" (the negative value next to the coverage) represents the percentage of newly added or modified lines of code that are not covered by unit tests. It calculates the difference in test coverage based solely on the changes made in the current commit or pull request. For example, if 10 lines of code are modified or added, and 8 of those lines are covered by unit tests, the "delta" would be -20%, indicating 20% of the newly changed code is untested. However, the "delta" has limitations. It can never be positive, meaning if you add more unit tests to cover existing, unmodified code, this additional coverage is not reflected in the delta. The metric only considers lines directly changed in the current set of modifications, as there is no mechanism to track improvements in coverage for pre-existing code that hasn't been altered.`

Do the numbers make more sense with this explanation?

salamonpavel:
> > > > Any chance to increase the coverage %, @salamonpavel ?
> > > 
> > > 
> > > Coincidentally @miroslavpojer today mentioned that jacoco-report had release version 1.7.1 with number of bugifxes. Can you please increase the version in the script to test and see how the new version works.
> > 
> > 
> > Those coverage numbers are incorrect.
> 
> The new jacoco-report version explains the meaning of the negative percentage numbers.
> 
> `The "delta" (the negative value next to the coverage) represents the percentage of newly added or modified lines of code that are not covered by unit tests. It calculates the difference in test coverage based solely on the changes made in the current commit or pull request. For example, if 10 lines of code are modified or added, and 8 of those lines are covered by unit tests, the "delta" would be -20%, indicating 20% of the newly changed code is untested. However, the "delta" has limitations. It can never be positive, meaning if you add more unit tests to cover existing, unmodified code, this additional coverage is not reflected in the delta. The metric only considers lines directly changed in the current set of modifications, as there is no mechanism to track improvements in coverage for pre-existing code that hasn't been altered.`
> 
> Do the numbers make more sense with this explanation? I still need the action control to switch this feature off.

Every single line in GetPartitioningCheckpoints.scala is invoked when testing the code yet it reports coverage of 58.1 %.

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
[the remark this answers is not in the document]
lsulak: Hmm what about function `flows.get_flow_checkpoints` ?

miroslavpojer:
> > > > > Any chance to increase the coverage %, @salamonpavel ?
> > > > 
> > > > 
> > > > Coincidentally @miroslavpojer today mentioned that jacoco-report had release version 1.7.1 with number of bugifxes. Can you please increase the version in the script to test and see how the new version works.
> > > 
> > > 
> > > Those coverage numbers are incorrect.
> > 
> > 
> > The new jacoco-report version explains the meaning of the negative percentage numbers.
> > `The "delta" (the negative value next to the coverage) represents the percentage of newly added or modified lines of code that are not covered by unit tests. It calculates the difference in test coverage based solely on the changes made in the current commit or pull request. For example, if 10 lines of code are modified or added, and 8 of those lines are covered by unit tests, the "delta" would be -20%, indicating 20% of the newly changed code is untested. However, the "delta" has limitations. It can never be positive, meaning if you add more unit tests to cover existing, unmodified code, this additional coverage is not reflected in the delta. The metric only considers lines directly changed in the current set of modifications, as there is no mechanism to track improvements in coverage for pre-existing code that hasn't been altered.`
> > Do the numbers make more sense with this explanation? I still need the action control to switch this feature off.
> 
> Every single line in GetPartitioningCheckpoints.scala is invoked when testing the code yet it reports coverage of 58.1 %.

The reason for 58,1% is in jacoco.xml
I have retested it on my local and there is 100% coverage after our scala filtering. I looks that scala filtering was not fully applied for XML file.

database/src/main/postgres/runs/V1.8.3__get_partitioning_checkpoints.sql
```diff
@@ -28,86 +28,113 @@ CREATE OR REPLACE FUNCTION runs.get_partitioning_checkpoints(
...
-            C.process_start_time,
-            C.id_checkpoint
-        LIMIT nullif(i_limit, 0);
+            LC.id_checkpoint, LC.process_start_time;
 
+    IF NOT FOUND THEN
+        status := 12;
+        status_text := 'OK with no checkpoints found';
+        RETURN NEXT;
+    END IF;
```
benedeki, line 134:
```suggestion
```

benedeki (approved):

PR #274, opened 2024-09-18
verify docker image

### LINKED ISSUE #260: Changing/verifying Atum docker definition
## Background
Steps to ensure Atum 0.3.0 release and deploy

## Feature
The docker definition might need an update after all the server changes. It can be, that it's up to date, then just verify it's OK.

salamonpavel:
Dockerfile and application configuration verified for upcoming deployment.
Closes #260

### DIFF
--- a/project/Setup.scala
+++ b/project/Setup.scala
@@ -55,7 +55,7 @@ object Setup {
   }
 
   val serverMergeStrategy = assembly / assemblyMergeStrategy := {
-    case PathList("META-INF", "services") => MergeStrategy.filterDistinctLines
+    case PathList("META-INF", "services", _*) => MergeStrategy.concat
     case PathList("META-INF", "maven", "org.webjars", "swagger-ui", "pom.properties") => MergeStrategy.singleOrError
     case PathList("META-INF", "resources", "webjars", "swagger-ui", _*) => MergeStrategy.singleOrError
     case PathList("META-INF", _*) => MergeStrategy.discard
--- a/server/src/main/resources/logback.xml
+++ b/server/src/main/resources/logback.xml
@@ -5,9 +5,9 @@
             <Pattern>%d{HH:mm:ss.SSS} %-5level [%thread] %logger{36}:%L - %msg %ex{short}%n</Pattern>
         </layout>
     </appender>
-    <root level="debug">
+    <root level="info">
         <appender-ref ref="STDOUT"/>
     </root>
-    <logger name="za.co.absa" level="debug"/>
+    <logger name="za.co.absa" level="info"/>
     <logger name="com.zaxxer.hikari" level="warn"/>
 </configuration>
--- a/server/src/main/resources/reference.conf
+++ b/server/src/main/resources/reference.conf
@@ -2,7 +2,8 @@
   postgres {
     # The JDBC driver class
     dataSourceClass=org.postgresql.Driver
-    serverName=localhost // host.docker.internal for local run in docker against db on its host machine; localhost otherwise for testing and for the gh pipeline
+    // host.docker.internal for local run against db running in docker on its host machine; localhost otherwise for testing and for the gh pipeline
+    serverName=localhost
     portNumber=5432
     databaseName=atum_db
     user=atum_user
@@ -16,8 +17,8 @@
   }
   ssl {
     enabled=false
-    keyStorePassword=password
-    keyStorePath="/path/to/your/cert"
+    keyStorePassword=changeit
+    keyStorePath="/etc/ssl/certs/selfsigned.jks"
   }
   monitoring {
     # monitoring of http communication
--- a/server/src/main/resources/reference.conf
+++ b/server/src/main/resources/reference.conf
@@ -2,7 +2,7 @@
   postgres {
     # The JDBC driver class
     dataSourceClass=org.postgresql.Driver
-    // host.docker.internal for local run against db running in docker on its host machine; localhost otherwise for testing and for the gh pipeline
+    # host.docker.internal for local run against db running in docker on its host machine; localhost otherwise for testing and for the gh pipeline
     serverName=localhost
     portNumber=5432
     databaseName=atum_db
@@ -18,6 +18,9 @@
   ssl {
     enabled=false
     keyStorePassword=changeit
+    # location of Java Keystore (JKS) file with certificate to be used for HTTPS communication;
+    # it can be sufficient to use self-signed certificate on testing or development environment or
+    # when application runs in an internal network; otherwise certificates are usually obtained from a trusted CA
     keyStorePath="/etc/ssl/certs/selfsigned.jks"
   }
   monitoring {
--- a/server/README.md
+++ b/server/README.md
@@ -48,3 +48,25 @@ When monitoring enabled, the application exposes `http://localhost:8080/metrics`
 and/or `http://localhost:8080/zio-metrics` endpoints which can be scraped by Prometheus.
 For testing purposes there is [docker-compose.yml](./docker-compose.yml) file which can be used to start up dockerized
 Prometheus and Grafana instances. Prometheus scraping configs are defined in [prometheus.yml](./prometheus.yml) file. 
+
+# Build & run docker image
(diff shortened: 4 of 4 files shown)


salamonpavel:
Release notes:

- Dockerfile and application configuration verified for deployment with ZIO and Http4s web server

project/Setup.scala
```diff
@@ -55,7 +55,7 @@ object Setup {
   }
 
   val serverMergeStrategy = assembly / assemblyMergeStrategy := {
-    case PathList("META-INF", "services") => MergeStrategy.filterDistinctLines
+    case PathList("META-INF", "services", _*) => MergeStrategy.concat
```
benedeki, line 58: What if there are same keys? (not sure what strategies there are)

server/src/main/resources/logback.xml
```diff
@@ -5,9 +5,9 @@
             <Pattern>%d{HH:mm:ss.SSS} %-5level [%thread] %logger{36}:%L - %msg %ex{short}%n</Pattern>
         </layout>
     </appender>
-    <root level="debug">
+    <root level="info">
```
benedeki, line 8: 👍

server/src/main/resources/reference.conf
```diff
@@ -16,8 +17,8 @@
   }
   ssl {
     enabled=false
-    keyStorePassword=password
-    keyStorePath="/path/to/your/cert"
+    keyStorePassword=changeit
```
benedeki, line 20: **Minor:**
Usually we use _changeme_ 😉

lsulak: `changeit` I think is the default pwd for the java cert store

salamonpavel: yup, changeit is default usually for cert store

server/src/main/resources/reference.conf
```diff
@@ -16,8 +17,8 @@
...
+    keyStorePath="/etc/ssl/certs/selfsigned.jks"
```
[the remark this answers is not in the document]
salamonpavel: @lsulak do you have any suggestion for a good comment here?

server/src/main/resources/reference.conf
benedeki: Sorry, did not know 😊

benedeki:
Seems OK to me, but don't feel competent in docker to judge the completeness.

project/Setup.scala
salamonpavel: In the context of the `sbt-assembly` plugin, merge strategies determine how to handle files with the same path when creating an assembly JAR. Here are the common merge strategies:

### `MergeStrategy.concat`
- **Purpose**: Concatenates the contents of all files with the same path.
- **Usage**: Useful when you want to combine the contents of multiple files into one, preserving all lines from each file.
- **Example**:
  ```scala
  case PathList("META-INF", "services", _*) => MergeStrategy.concat
  ```

### `MergeStrategy.singleOrError`
- **Purpose**: Ensures that only one file is included in the final assembly. If multiple files with the same path are found, it throws an error.
- **Usage**: Useful when you need to guarantee that there is no ambiguity or conflict for a particular file.
- **Example**:
  ```scala
  case PathList("META-INF", "maven", "org.webjars", "swagger-ui", "pom.properties") => MergeStrategy.singleOrError
  ```

### `MergeStrategy.rename`
- **Purpose**: Handles file conflicts by renaming the conflicting files.
- **Usage**: Useful when you want to include all versions of a file without merging their contents.
- **Example**:
  ```scala
  case PathList("META-INF", "services", _*) => MergeStrategy.rename
  ```

### `MergeStrategy.discard`
- **Purpose**: Discards all files with the same path.
- **Usage**: Useful when you want to exclude certain files from the final assembly.
- **Example**:
  ```scala
  case PathList("META-INF", _*) => MergeStrategy.discard
  ```

### `MergeStrategy.first`
- **Purpose**: Keeps the first file encountered and discards the rest.
- **Usage**: Useful when the order of files is known and you only need the first one.
- **Example**:
  ```scala
  case _ => MergeStrategy.first
  ```

### `MergeStrategy.last`
- **Purpose**: Keeps the last file encountered and discards the rest.
- **Usage**: Useful when the order of files is known and you only need the last one.
- **Example**:
  ```scala
  case _ => MergeStrategy.last
  ```

### `MergeStrategy.filterDistinctLines`
- **Purpose**: Merges files by including only distinct lines.
- **Usage**: Useful when you want to avoid duplicate lines in the merged file.
- **Example**:
  ```scala
  case "application.conf" => MergeStrategy.filterDistinctLines
  ```

### `MergeStrategy.deduplicate`
- **Purpose**: Merges files by including only distinct files.
- **Usage**: Useful when you want to avoid duplicate files in the merged file.
- **Example**:
  ```scala
  case "reference.conf" => MergeStrategy.deduplicate
  ```

### Custom Merge Strategy
- **Purpose**: Allows you to define custom logic for merging files.
- **Usage**: Useful when none of the predefined strategies fit your needs.
- **Example**:
  ```scala
  case "custom.conf" => new MergeStrategy {
    override def name: String = "customMergeStrategy"
    override def apply(tmp: File, path: String, files: Seq[File]): Either[String, Seq[File]] = {
      // Custom merge logic here
      Right(files)
    }
  }
  ```

These strategies help manage how files are handled during the assembly process, ensuring that the final JAR is built according to your project's requirements.

salamonpavel: `MergeStrategy.filterDistinctLines` and `MergeStrategy.concat` serve different purposes, and the choice between them depends on your specific requirements.

### `MergeStrategy.filterDistinctLines`
- **Purpose**: Merges files by including only distinct lines.
- **Usage**: Useful when you want to avoid duplicate lines in the merged file.
- **Safety**: Safer in scenarios where duplicate lines could cause issues, as it ensures each line appears only once.

### `MergeStrategy.concat`
- **Purpose**: Concatenates the contents of all files with the same path.
- **Usage**: Useful when you want to combine the contents of multiple files into one, preserving all lines from each file.
- **Safety**: May lead to conflicts or unexpected behavior if there are lines with the same keys but different values.

### Conclusion
- **`filterDistinctLines`** is generally safer when you need to ensure that there are no duplicate lines in the merged file.
- **`concat`** is useful when you need to preserve all lines, but it requires careful handling of potential conflicts.

lsulak: Thank you very much for the educative window! 

I can't tell you my preference because I don't know what duplicates we have here, and what lines. So I will base my answer according to these details - were you able to find out the specifics? 

Also, was there a problem so that you changed it, or you just experimented with it? At first, I thought that `filterDistinctLines` was not working for us, that's why you changed it?

server/src/main/resources/reference.conf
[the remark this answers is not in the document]
lsulak: not sure, but perhaps something like this?

// location of Java Keystore (JKS) file with certificate to be used for HTTPS communication; it can be sufficient to use self-signed certificate on testing or development environment or when application runs in an internal network, otherwise certificates are usually obtained from a trusted CA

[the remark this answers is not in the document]
salamonpavel: it's a bit wordy but thanks will use it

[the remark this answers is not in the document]
lsulak: yeah I know, feel free to change it :D

project/Setup.scala
salamonpavel: The change wasn't just about the strategy. Honestly I didn't dig very deeply into this, just made it work by trial/error. At this moment I would rather dedicate my time to something more pressing than exploring this further.

lsulak: Understood, well then I don't really have preferences. Whatever works would be my answer, and if we encounter a problem around these, we can revisit it later with this knowledge ^ in mind 🤷

lsulak (approved):
Looks good, I like that you also added the docker build & run command into the README with all the parameters we talked about - I didn't try it yet on my machine but we did it together on a call. Approved, good job :)

Issue #263, opened 2024-09-04
Hide non-active endpoints (v2) from swagger to make them silent live


benedeki:
## Background
Steps to ensure Atum 0.3.0 release and deploy.
Some things are not directly used yet in the parts of the project. To be able to change them, in case the need would arise without the hassle of maintaining possible compatibility, we make them truly "silent live" by not even documenting such end-points.

## Feature
Comment out unused (in _Agent_) REST API endpoints, so they don't appear in Swagger.

PR #277, opened 2024-09-25
Hiding in-active endpoints: #263

TebaleloS:
# Release notes

* This PR implement the functionality if hiding in-active endpoint from swagger documentation.

`Closes` #263

### DIFF
--- a/server/src/main/scala/za/co/absa/atum/server/api/http/Routes.scala
+++ b/server/src/main/scala/za/co/absa/atum/server/api/http/Routes.scala
@@ -114,17 +114,17 @@ trait Routes extends Endpoints with ServerOptions {
   private def createSwaggerRoutes: HttpRoutes[HttpEnv.F] = {
     val endpoints = List(
       createCheckpointEndpointV1,
-      postCheckpointEndpointV2,
+      //      postCheckpointEndpointV2,
       createPartitioningEndpointV1,
-      postPartitioningEndpointV2,
-      patchPartitioningAdditionalDataEndpointV2,
-      getPartitioningCheckpointsEndpointV2,
-      getPartitioningCheckpointEndpointV2,
-      getPartitioningMeasuresEndpointV2,
-      getPartitioningEndpointV2,
-      getPartitioningMeasuresEndpointV2,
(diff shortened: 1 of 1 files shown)


TebaleloS:
# Release notes

* This PR implement the functionality if hiding in-active endpoint from swagger documentation.

server/src/main/scala/za/co/absa/atum/server/api/http/Routes.scala
```diff
@@ -95,14 +95,14 @@ trait Routes extends Endpoints with ServerOptions {
   private def createSwaggerRoutes: HttpRoutes[HttpEnv.F] = {
     val endpoints = List(
       createCheckpointEndpointV1,
-      postCheckpointEndpointV2,
+//      postCheckpointEndpointV2,
       createPartitioningEndpointV1,
-      postPartitioningEndpointV2,
+//      postPartitioningEndpointV2,
       patchPartitioningAdditionalDataEndpointV2,
```
benedeki, line 101: Do you guys know, if we use this endpoint in Agent now? @salamonpavel @lsulak

benedeki (commented):
Looks good.
Just to solve the questions.
Mind that in case of conflicts (other endpoints merged) the list can be edited on either side.

server/src/main/scala/za/co/absa/atum/server/api/http/Routes.scala
lsulak: I don't think so, PATCH wasn't introduced into the agent.

Agent is using this endpoint: `POST /api/v2/writeAdditionalData`

lsulak: actually perhaps we should adjust the agent to work with this PATCH because otherwise we got incompatibility in 0.3.0 between agent and server

benedeki (dismissed):

TebaleloS:
# Release notes

* This PR implement the functionality if hiding in-active endpoint from swagger documentation.

benedeki (approved):

PR #283, opened 2024-10-01
#280: agent and server compatibility for Additional Data PATCH operation

### LINKED ISSUE #280: Create Additional Data - adopt agent & test Agent <-> Server communication
## Background
As the title says.

## Feature
Atum Agent fully supports Additional Data creation.

lsulak:
Note: Additional Data = AD

What was done:
* a lot more than I wanted :( (but most files changed are a result of moving files from server/ to model/ - REST envelope classes)
* AD update functionality compatible between Agent <-> Server
* Changing HTTP Backend on Agent from `HttpURLConnectionBackend` -> `OkHttpSyncBackend` - inspired by https://github.com/softwaremill/sttp/issues/316 and https://github.com/raster-foundry/raster-foundry/pull/5306 because `HttpURLConnectionBackend` didn't support PATCH HTTP Method and we need it for AD update endpoint
* Creating a general DTO Base 64 Encode function + tests
* Moved Envelope related code from Server to Model, because Agent also needs it
* Added E2E test using Balta to check the results - this is a new 'type' of test and will not be executed in CI yet. But if your local PG is ready & empty and the service is running locally, it will work with the proper service configuration 
* Removed a bunch of obsolete AD DTOs
* Server supports these 2 endpoints: `patchPartitioningAdditionalDataEndpointV2` and `getPartitioningEndpointV2`

Note: If you want to run `AgentWithServerIntegrationTests` then:
* Prepare PG locally (make sure it has newest schemas / tables, and that it's empty)
* Prepare Service - either as Dockerized one or a typical Localhost - and adjust `src/test/resources/reference.conf` accordingly - this is needed: `atum.dispatcher.http.url="http://localhost:8080"` and `atum.dispatcher.type="http"`
* I personally run PG in Docker & Service in Docker and it all works very nicely

Closes #280


lsulak:
Release Notes:
* Support of Additional Data Update functionality

benedeki (dismissed):
Reviewed on a pair session

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAdditionalDataIntegrationTests.scala
```diff
@@ -88,19 +88,21 @@ class GetPartitioningAdditionalDataIntegrationTests extends DBTestSuite{
     )
 
     function(fncGetPartitioningAdditionalData)
-      .setParam("i_partitioning", partitioning1)
+      .setParam("i_partitioning_id", fkPartitioning1)
```
benedeki, line 91: You know, that you don't have to name the params, you can put them there positionally? Especially in one param functions, it's easier

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAdditionalDataIntegrationTests.scala
```diff
@@ -88,19 +88,21 @@ class GetPartitioningAdditionalDataIntegrationTests extends DBTestSuite{
...
       .execute { queryResult =>
         val results = queryResult.next()
         assert(results.getInt("status").contains(11))
         assert(results.getString("status_text").contains("OK"))
         assert(results.getString("ad_name").contains("ad_1"))
         assert(results.getString("ad_value").contains("This is the additional data for Joseph"))
+        assert(results.getString("ad_author").contains("Joseph"))
```
benedeki, line 98: Really need to switch to Balta 0.3, so much tidier 😉

benedeki (approved):
LGTM

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAdditionalDataIntegrationTests.scala
lsulak: Yep, I know, but I find it nicer - more readable if it's named, especially if there are more functions; but for 1-2 params I think you are right, can be ommited

PR #311, opened 2025-01-21
Added the getAncestors Database functionality.

ABLL526:
This PR is strictly for the getAncestors DataBase Functionality.
Any feedback is welcome.

### DIFF
--- a/database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
+++ b/database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION runs.get_ancestors(
+    IN i_id_partitioning    BIGINT,
+    IN i_limit              INT DEFAULT 5,
+    IN i_offset             BIGINT DEFAULT 0,
+    OUT status              INTEGER,
+    OUT status_text         TEXT,
+    OUT ancestorid         BIGINT,
+    OUT partitioning        JSONB,
+    OUT author              TEXT,
+    OUT has_more            BOOLEAN
+)  RETURNS SETOF record AS
+$$
+    -------------------------------------------------------------------------------
+--
+-- Function: runs.get_ancestors(3)
+--      Returns Ancestors' partition ID for the given id
+--
+-- Parameters:
+--      i_id_partitioning   - id that we asking the Ancestors for
+--      i_limit             - (optional) maximum number of partitionings to return, default is 5
+--      i_offset            - (optional) offset to use for pagination, default is 0
+--
+-- Returns:
+--      status              - Status code
+--      status_text         - Status message
+--      ancestorid         - ID of Ancestor partition
+--      partitioning        - partitioning data of ancestor
+--      author              - author of the Ancestor partitioning
+--      has_more            - Flag indicating if there are more partitionings available
+
+-- Status codes:
+--      11 - OK
+--      41 - Partitioning not found
+--      42 - Ancestor Partitioning not found
+--
+-------------------------------------------------------------------------------
+DECLARE
+    partitionCreateAt TIMESTAMP;
+    _has_more BOOLEAN;
+
+BEGIN
+    -- Check if the partitioning exists
+    PERFORM 1 FROM runs.partitionings WHERE id_partitioning = i_id_partitioning;
+    IF NOT FOUND THEN
+        status := 41;
+        status_text := 'Partitioning not found';
+        RETURN NEXT;
+        RETURN;
+    END IF;
+
+    -- Get the creation timestamp of the partitioning
+    SELECT created_at
+    FROM runs.partitionings
+    WHERE id_partitioning = i_id_partitioning
+    INTO partitionCreateAt;
+
+    -- Check if there are more partitionings than the limit
+    SELECT count(*) > i_limit
+    FROM flows.partitioning_to_flow PTF
+    WHERE PTF.fk_flow IN (
+        SELECT fk_flow
+        FROM flows.partitioning_to_flow
+        WHERE fk_partitioning = i_id_partitioning
+    )
+    LIMIT i_limit + 1 OFFSET i_offset
+    INTO _has_more;
+
+    -- Return the ancestors
+    RETURN QUERY
+        SELECT
+            11 AS status,
+            'OK' AS status_text,
+            P.id_partitioning AS ancestorid,
+            P.partitioning AS partitioning,
+            P.created_by AS author,
+            _has_more AS has_more
+        FROM
+            runs.partitionings P
+                INNER JOIN flows.partitioning_to_flow PF ON PF.fk_partitioning = P.id_partitioning
+                INNER JOIN flows.partitioning_to_flow PF2 ON PF2.fk_flow = PF.fk_flow
+        WHERE
+            PF2.fk_partitioning = i_id_partitioning
+            AND
+            P.created_at < partitionCreateAt
+        GROUP BY P.id_partitioning
+        ORDER BY P.id_partitioning, P.created_at DESC
+        LIMIT i_limit
+        OFFSET i_offset;
+
+    IF FOUND THEN
+        status := 11;
+        status_text := 'OK';
+    ELSE
+        status := 42;
+        status_text := 'Ancestor Partitioning not found';
+    END IF;
+    RETURN NEXT;
+    RETURN;
+
+END;
+$$
+    LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
+
+ALTER FUNCTION runs.get_ancestors(BIGINT, INT, BIGINT) OWNER TO atum_owner;
+GRANT EXECUTE ON FUNCTION runs.get_ancestors(BIGINT, INT, BIGINT) TO atum_user;
--- a/database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
+++ b/database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
@@ -1,0 +1,533 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.database.runs
+
+import io.circe.parser.parse
+import za.co.absa.balta.DBTestSuite
+import za.co.absa.balta.classes.JsonBString
+
+import java.time.OffsetDateTime
+
+class GetAncestorsIntegrationTests extends DBTestSuite {
+
+  private val getAncestorsFn = "runs.get_ancestors"
+  private val partitioningsTable = "runs.partitionings"
+  private val createPartitioningFn = "runs.create_partitioning"
+
+  private val createFlowFn = "flows._create_flow"
+  private val addToParentFlowsFn = "flows._add_to_parent_flows"
+
+  private val partitioning1 = JsonBString(
+    """
+      |{
+      |   "version": 1,
+      |   "keys": ["keyA", "keyB", "keyC"],
+      |   "keysToValuesMap": {
+      |     "keyA": "valueA",
+      |     "keyB": "valueB",
+      |     "keyC": "valueC"
+      |   }
+      |}
+      |""".stripMargin
+  )
+
+  private val expectedPartitioning1 = parse(partitioning1.value).getOrElse(throw new Exception("Failed to parse JSON"))
+
+  private val partitioning2 = JsonBString(
+    """
+      |{
+      |   "version": 1,
+      |   "keys": ["keyD", "keyE", "keyF"],
+      |   "keysToValuesMap": {
+      |     "keyD": "valueD",
+      |     "keyE": "valueE",
+      |     "keyF": "valueF"
+      |   }
+      |}
+      |""".stripMargin
+  )
+
+  private val expectedPartitioning2 = parse(partitioning2.value).getOrElse(throw new Exception("Failed to parse JSON"))
+
+  private val partitioning3 = JsonBString(
+    """
+      |{
+      |   "version": 1,
+      |   "keys": ["keyG", "keyH", "keyI"],
+      |   "keysToValuesMap": {
+      |     "keyG": "valueG",
+      |     "keyH": "valueH",
+      |     "keyI": "valueI"
+      |   }
+      |}
+      |""".stripMargin
+  )
+
+  private val expectedPartitioning3 = parse(partitioning3.value).getOrElse(throw new Exception("Failed to parse JSON"))
+
+  private val partitioning4 = JsonBString(
+    """
+      |{
+      |   "version": 1,
+      |   "keys": ["keyJ", "keyK", "keyL"],
+      |   "keysToValuesMap": {
+      |     "keyJ": "valueJ",
+      |     "keyK": "valueK",
+      |     "keyL": "valueL"
+      |   }
+      |}
+      |""".stripMargin
+  )
+
+  private val expectedPartitioning4 = parse(partitioning4.value).getOrElse(throw new Exception("Failed to parse JSON"))
+
+  private val partitioning5 = JsonBString(
+    """
+      |{
+      |   "version": 1,
+      |   "keys": ["keyM", "keyN", "keyO"],
+      |   "keysToValuesMap": {
+      |     "keyM": "valueM",
+      |     "keyN": "valueN",
+      |     "keyO": "valueO"
+      |   }
+      |}
+      |""".stripMargin
+  )
+
+  private val expectedPartitioning5 = parse(partitioning5.value).getOrElse(throw new Exception("Failed to parse JSON"))
+
+  private val partitioning6 = JsonBString(
+    """
+      |{
+      |   "version": 1,
+      |   "keys": ["keyP", "keyQ", "keyR"],
+      |   "keysToValuesMap": {
+      |     "keyP": "valueP",
+      |     "keyQ": "valueQ",
+      |     "keyR": "valueR"
+      |   }
+      |}
+      |""".stripMargin
+  )
+
+  private val expectedPartitioning6 = parse(partitioning6.value).getOrElse(throw new Exception("Failed to parse JSON"))
+
+  private val partitioning7 = JsonBString(
+    """
+      |{
+      |   "version": 1,
+      |   "keys": ["keyS", "keyT", "keyU"],
+      |   "keysToValuesMap": {
+      |     "keyS": "valueS",
+      |     "keyT": "valueT",
+      |     "keyU": "valueU"
+      |   }
+      |}
+      |""".stripMargin
+  )
+
+  private val expectedPartitioning7 = parse(partitioning7.value).getOrElse(throw new Exception("Failed to parse JSON"))
+
+  private val partitioning8 = JsonBString(
+    """
+      |{
+      |   "version": 1,
+      |   "keys": ["keyV", "keyW", "keyX"],
+      |   "keysToValuesMap": {
+      |     "keyV": "valueV",
+      |     "keyW": "valueW",
+      |     "keyX": "valueX"
+      |   }
+      |}
+      |""".stripMargin
+  )
+
+  var flowIdOfPartitioning1: Long = _
+  var flowIdOfPartitioning2: Long = _
+  var flowIdOfPartitioning3: Long = _
+
+  test("Returns Ancestors for a given Partition ID") {
+
+    val Time1 = OffsetDateTime.parse("1992-08-03T10:00:00Z")
+    val Time2 = OffsetDateTime.parse("1992-08-04T10:00:00Z")
+    val Time3 = OffsetDateTime.parse("1992-08-05T10:00:00Z")
+    val Time4 = OffsetDateTime.parse("1992-08-06T10:00:00Z")
+    val Time5 = OffsetDateTime.parse("1992-08-07T10:00:00Z")
+    val Time6 = OffsetDateTime.parse("1992-08-08T10:00:00Z")
+    val Time7 = OffsetDateTime.parse("1992-08-09T10:00:00Z")
+    val Time8 = OffsetDateTime.parse("1992-08-09T11:00:00Z")
+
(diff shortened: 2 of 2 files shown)


database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
```diff
@@ -0,0 +1,121 @@
...
+-- Returns:
+--      status              - Status code
+--      status_text         - Status message
+--      ancestorid         - ID of Ancestor partition
+--      partitioning        - partitioning data of ancestor
+--      author              - author of the Ancestor partitioning
+--      has_more            - Flag indicating if there are more partitionings available
+
+-- Status codes:
+--      11 - OK
```
salamonpavel, line 48: Not super important I suppose but still, according to https://github.com/AbsaOSS/fa-db/blob/master/core/src/main/scala/za/co/absa/db/fadb/status/README.md we would maybe want to use status 10 instead of 11.

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
```diff
@@ -0,0 +1,121 @@
...
+    OUT ancestorid         BIGINT,
+    OUT partitioning        JSONB,
+    OUT author              TEXT,
+    OUT has_more            BOOLEAN
+)  RETURNS SETOF record AS
+$$
+    -------------------------------------------------------------------------------
+--
+-- Function: runs.get_ancestors(3)
+--      Returns Ancestors' partition ID for the given id
```
salamonpavel, line 32: Let's not introduce confusion. In Atum context 'partitioning' and 'partition' are different terms. Partitioning consists of set of partitions. Moreover the function returns only only the ancestor's id.

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
```diff
@@ -0,0 +1,121 @@
...
+--      has_more            - Flag indicating if there are more partitionings available
+
+-- Status codes:
+--      11 - OK
+--      41 - Partitioning not found
+--      42 - Ancestor Partitioning not found
+--
+-------------------------------------------------------------------------------
+DECLARE
+    partitionCreateAt TIMESTAMP;
```
salamonpavel, line 54: partitioning

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
```diff
@@ -0,0 +1,121 @@
...
+        WHERE
+            PF2.fk_partitioning = i_id_partitioning
+            AND
+            P.created_at < partitionCreateAt
+        GROUP BY P.id_partitioning
+        ORDER BY P.id_partitioning, P.created_at DESC
+        LIMIT i_limit
+        OFFSET i_offset;
+
+    IF FOUND THEN
```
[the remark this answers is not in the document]
salamonpavel: We then simply process the data as
```scala
if (results.nonEmpty && results.head.hasMore) ...
```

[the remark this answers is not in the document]
ABLL526: This makes sense although. From runs.get_paritioning_checkpoint_v2 it has a similar logic to this. 
What I will do is comment it out for now and determine if it is necessary.

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
ABLL526: Changed as mentioned

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
ABLL526: I will make the change throughout the SQL for consistency. Thanks, Pavel

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
```diff
@@ -0,0 +1,121 @@
...
+            P.created_by AS author,
+            _has_more AS has_more
+        FROM
+            runs.partitionings P
+                INNER JOIN flows.partitioning_to_flow PF ON PF.fk_partitioning = P.id_partitioning
+                INNER JOIN flows.partitioning_to_flow PF2 ON PF2.fk_flow = PF.fk_flow
+        WHERE
+            PF2.fk_partitioning = i_id_partitioning
+            AND
+            P.created_at < partitionCreateAt
```
benedeki, line 100: Why this condition?
Actually I think the whole query is incorrect, unfortunately.
It should be
```sql
FROM
            flows.partitioning_to_flow PF
                INNER JOIN flows.flows F ON F.id_flow = PF.id_flow
                INNER JOIN runs.partitionings P ON P.id_partitioning = F.fk_primary_partitioning
        WHERE
            PF.fk_partitioning = i_id_partitioning AND
            P.id_partitioning IS DISTINCT FROM i_id_partitioning 

```

benedeki (changes requested):

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
```diff
@@ -0,0 +1,121 @@
...
+)  RETURNS SETOF record AS
+$$
+    -------------------------------------------------------------------------------
+--
+-- Function: runs.get_ancestors(3)
+--      Returns Ancestors' partition ID for the given id
+--
+-- Parameters:
+--      i_id_partitioning   - id that we asking the Ancestors for
+--      i_limit             - (optional) maximum number of partitionings to return, default is 5
```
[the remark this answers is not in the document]
ABLL526: I will change it to 10. I don't think it is important either.

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
ABLL526: Thank you the convention is useful to know. I have made the change.

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
ABLL526: I understand, but I think that it should be as the current since your query would get the children partitionings as well. Not just the Ancestors. Please let me know if I am mistaken with my analysis on this.

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
```diff
@@ -0,0 +1,121 @@
...
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION runs.get_ancestors(
+    IN i_id_partitioning    BIGINT,
+    IN i_limit              INT DEFAULT 5,
+    IN i_offset             BIGINT DEFAULT 0,
+    OUT status              INTEGER,
+    OUT status_text         TEXT,
+    OUT ancestorid         BIGINT,
```
lsulak, line 23:
```suggestion
    OUT ancestor_id         BIGINT,
```

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
```diff
@@ -0,0 +1,121 @@
...
+--      status_text         - Status message
+--      ancestorid         - ID of Ancestor partition
+--      partitioning        - partitioning data of ancestor
+--      author              - author of the Ancestor partitioning
+--      has_more            - Flag indicating if there are more partitionings available
+
+-- Status codes:
+--      11 - OK
+--      41 - Partitioning not found
+--      42 - Ancestor Partitioning not found
```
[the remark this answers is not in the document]
lsulak: I agree

database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,533 @@
...
+        val row = queryResult.next()
+        val returnedPartitioning = row.getJsonB("partitioning").get
+        val returnedPartitioningParsed = parse(returnedPartitioning.value)
+          .getOrElse(fail("Failed to parse returned partitioning"))
+        assert(row.getInt("status").contains(11))
+        assert(row.getString("status_text").contains("OK"))
+        assert(row.getLong("ancestorid").contains(partId1))
+        assert(returnedPartitioningParsed == expectedPartitioning1)
+        assert(row.getString("author").contains("Grandpa"))
+        assert(row.getString("author").contains("Grandpa"))
```
lsulak, line 312: also I'd recommend to test whether the result is final, by adding this on the end of this local scope:

`assert(!queryResult.next())

database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,533 @@
...
+        assert(row.getString("author").contains("Son"))
+        row = queryResult.next()
+        returnedPartitioning = row.getJsonB("partitioning").get
+        returnedPartitioningParsed = parse(returnedPartitioning.value)
+          .getOrElse(fail("Failed to parse returned partitioning"))
+        assert(row.getInt("status").contains(11))
+        assert(row.getString("status_text").contains("OK"))
+        assert(row.getLong("ancestorid").contains(partId4))
+        assert(returnedPartitioningParsed == expectedPartitioning4)
+        assert(row.getString("author").contains("Grandson"))
```
lsulak, line 354: again, add the check on finality

Btw, I draw a simple diagram of partitioning hierarchy from this test and I see that this test checks ancestors correctly. For other reviewers, this is the hierarchy - ancestors are requested for Partitioning 5, and ancestors are: 1, 2, 3, 4:

```
 1 2
 | | 
 3 4  6
 \ |  |
   5  7
   | /
   8
```

consider adding this 'diagram' into the code comment somewhere for easier understanding

lsulak: also, you can use `.map` to iterate over a list of expected results - you are checking 4 rows and the code is almost the same, so can be massively simplified :)

database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,533 @@
...
+        assert(row.getString("author").contains("Mother"))
+        row = queryResult.next()
+        returnedPartitioning = row.getJsonB("partitioning").get
+        returnedPartitioningParsed = parse(returnedPartitioning.value)
+          .getOrElse(fail("Failed to parse returned partitioning"))
+        assert(row.getInt("status").contains(11))
+        assert(row.getString("status_text").contains("OK"))
+        assert(row.getLong("ancestorid").contains(partId7))
+        assert(returnedPartitioningParsed == expectedPartitioning7)
+        assert(row.getString("author").contains("Daughter"))
```
lsulak, line 439: same here - briefly checked, what you are checking seems to be correct, but can be massively simplified & I'd like you to add the last assert on data finality

database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,533 @@
...
+  test("Returns Ancestors for a given Partition ID") {
+
+    val Time1 = OffsetDateTime.parse("1992-08-03T10:00:00Z")
+    val Time2 = OffsetDateTime.parse("1992-08-04T10:00:00Z")
+    val Time3 = OffsetDateTime.parse("1992-08-05T10:00:00Z")
+    val Time4 = OffsetDateTime.parse("1992-08-06T10:00:00Z")
+    val Time5 = OffsetDateTime.parse("1992-08-07T10:00:00Z")
+    val Time6 = OffsetDateTime.parse("1992-08-08T10:00:00Z")
+    val Time7 = OffsetDateTime.parse("1992-08-09T10:00:00Z")
+    val Time8 = OffsetDateTime.parse("1992-08-09T11:00:00Z")
```
lsulak, line 173: I would propose to change these times so that they are not incremental - let's say that they are more 'random', let's say that there isn't this nice chronological sequence of partitioning time creation. For example, in real life, parent can be created in later moment than a child.

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
lsulak: It seems that you understood the problem, technologies, and I checked your tests and it looks logically correct. 

However, the biggest problem I see is this condition: `P.created_at < partitionCreateAt` - we cannot count on creation time - parent can be created before or after child, so you need to adjust the query

database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
ABLL526: Removing the entire premise of time creation from the query

database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
ABLL526: Thank you will modify once I have the solution correct. I have also added the diagram in the comments.

database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
ABLL526: Done

database/src/test/scala/za/co/absa/atum/database/runs/GetAncestorsIntegrationTests.scala
lsulak: Awesome, looking forward to it, I'll review once adjusted :)

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
ABLL526: Changed as needed

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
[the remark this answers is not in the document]
ABLL526: Removed as needed

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
ABLL526: I have made this change as discussed, I did not realise there was an added field in the Flows table. That made this very easy. Thank you for the feedback, it has been changed and everything is working as intended.

database/src/main/postgres/runs/V0.3.0.2__get_ancestors.sql
ABLL526: changed

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,425 @@
...
+
+  // Testing for return of the Ancestors for a given Partition ID
+  //
+  //  1(Grandma)  2(Grandpa)
+  //      |           |
+  //  3(Mother)   4(Father)    6(Daughter)
+  //     \        |                |
+  //       5(Son)           7(Granddaughter)
+  //          |            /
+  //           8(Grandson)
```
benedeki, line 213: Love the diagram 😎

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,425 @@
...
+  test("Returns Ancestors for a given Partition ID"){
+    val partitioningID1 = function(createPartitioningFn)
+      .setParam("i_partitioning", partitioning1)
+      .setParam("i_by_user", "Grandma")
```
benedeki, line 217: Putting the "family tree position" here is smart too. 👍

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,425 @@
...
+    //TEST ALL flows for Ancestors Partitions
+    function(getAncestorsFn)
+      .setParam("i_id_partitioning", partitioningID8)
+      .execute { queryResult =>
+        var row = queryResult.next()
+        var returnedPartitioning = row.getJsonB("partitioning").get
+        var returnedPartitioningParsed = parse(returnedPartitioning.value)
+          .getOrElse(fail("Failed to parse returned partitioning"))
+        //Used breakable to be able to break the loop
+        breakable {
```
benedeki, line 407: Not sure I understand the need for this.

database/src/main/postgres/runs/V0.3.0.2__get_partitioning_ancestors.sql
```diff
@@ -0,0 +1,111 @@
...
+        ORDER BY P.id_partitioning
+        LIMIT i_limit
+        OFFSET i_offset;
+
+    --If no ancestors found send an OK status
+    IF NOT FOUND THEN
+        status := 10;
+        status_text := 'OK';
+        RETURN NEXT;
+    END IF;
```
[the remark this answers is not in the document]
ABLL526: I will check the documentation to return an unexpected code. Or an empty answer.

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,425 @@
...
+    val nonExistentID = 9999L
+
+    function(getAncestorsFn)
+      .setParam("i_id_partitioning", nonExistentID)
+      .execute { queryResult =>
+        assert(queryResult.hasNext)
+        val row = queryResult.next()
+        assert(row.getInt("status").contains(41))
+        assert(row.getString("status_text").contains("Partitioning not found"))
+        assert(row.getJsonB("ancestor_id").isEmpty)
```
ABLL526, line 170: Testing null values is fine for test purposes, in reality it can have data. Therefore just testing the error codes is good enough.

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,425 @@
...
+        assert(row.getJsonB("partitioning").isEmpty)
+        assert(row.getString("author").isEmpty)
+        assert(!queryResult.hasNext)
+      }
+  }
+
+  //Second Failure Test: Ancestor Partitioning not found
+  test("Ancestor Partitioning not found") {
+
+    val partitioningID1 = function(createPartitioningFn)
```
ABLL526, line 180: Just call the function. Make the assumption it works.

ABLL526: Having the data preparation in one function.

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,391 @@
...
+        var returnedPartitioning = row.getJsonB("partitioning").get
+        var returnedPartitioningParsed = parse(returnedPartitioning.value)
+          .getOrElse(fail("Failed to parse returned partitioning"))
+        //Used breakable to be able to break the loop
+        breakable {
+          for ((k, v) <- resultsMap) {
+            assert(row.getInt("status").contains(10))
+            assert(row.getString("status_text").contains("OK"))
+            assert(row.getLong("ancestor_id").contains(v._1))
+            assert(returnedPartitioningParsed == v._2)
```
lsulak, line 298: you can probably expand `v` as pattern match instead of `._1` and `._2`, just an idea

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,391 @@
...
+            assert(row.getString("author").contains(k))
+            if (!queryResult.hasNext) break()
```
lsulak, line 300: this is a very non-functional, procedural-like code, which is, no matter the style, very non-deterministic. I would prepare exact expected parents and loop over them, not like this

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,391 @@
...
+            assert(queryResult.hasNext)
+            row = queryResult.next()
+            returnedPartitioning = row.getJsonB("partitioning").get
+            returnedPartitioningParsed = parse(returnedPartitioning.value)
```
lsulak, line 304: also here you are not really checking the expected & returned ancestors. This test basically doesn't test whether all expected ancestors are being returned, it really just checks P1, then if the query didn't return anything it breaks from the check loop. 

So this test doesn't prove that your DB function really works and therefore it really needs improvement

lsulak: Having non-determinism in tests and having partial-only expected results is definitely not a good practice

lsulak: the last 2 tests are OK, but this one seriously needs improvement

ABLL526: Thank you. I have removed all the breakable loops. I have then added a set case not partial-only expected results. Please check if this is a better practice.

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
ABLL526: Removed the breakable test

benedeki (approved):
By returning 14, you IMHO chose the more complicated path, but it's a valid one 😉

database/src/main/postgres/runs/V0.3.0.3__get_partitioning_ancestors.sql
```diff
@@ -0,0 +1,112 @@
...
+        WHERE
+            PF.fk_partitioning = i_id_partitioning AND
+            P.id_partitioning IS DISTINCT FROM i_id_partitioning
+        ORDER BY P.id_partitioning
+        LIMIT i_limit
+        OFFSET i_offset;
+
+    --If no ancestors found send an OK status
+    IF NOT FOUND THEN
+        status := 14;
```
lsulak, line 101: I don't know if this is gonna be useful, but as you said on some place above, function `runs.get_partitioning_checkpoint_v2` basically uses similar principle, so I guess you can keep it as is

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
```diff
@@ -0,0 +1,395 @@
...
+        assert(returnedPartitioningParsed == expectedPartitioning1)
+        assert(row.getString("author").contains("Grandma"))
+        assert(!queryResult.hasNext)
+      }
+
+    //Test Multiple Ancestors
+    function(getAncestorsFn)
+      .setParam("i_id_partitioning", partitioningID5)
+      .execute { queryResult =>
+        for ((k, v) <- resultsMapFor5) {
```
lsulak, line 297: You could perhaps do something like this, but I don't insist

`for ((k, (expectedPartitioningId, expectedPartitioning)) <- resultsMapFor5) {

lsulak (approved):
LGTM, apologies it took me so long to re-review, I love the changes, diagram, and the way you refactored it. One potential improvement above, but I don't insist - feel free to merge if you want :)

database/src/test/scala/za/co/absa/atum/database/runs/GetPartitioningAncestorsIntegrationTests.scala
ABLL526: Thank you.

I do like this better. Since the other PR is dependent on this one. I will make the change in the other PR.

Issue #247, opened 2024-08-19
Implement basics of `FlowReader`


benedeki:
## Background
The initial implementation of the _Reader_ module.

## Feature
Implement the `FlowReader` class. Its constructor (based on _primary partitioning_, and basic functions returning information on the `Flow`:
* partiotionings
* checkpoints

Depends on #245

benedeki:
This item depends on:
* AbsaOSS/atum-service#245
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

benedeki:
This item depends on:
* ~~AbsaOSS/atum-service#245~~
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

PR #316, opened 2025-02-04
#247: Paging support in Reader module

benedeki:
* classes abstracting work with paging and rolling through them
* `RequestResult` class refactoring
* package `basic` renamed to `core`

Requirement for #247

Release notes:
* Added paging support to the module.

### DIFF
--- a/reader/src/main/scala/za/co/absa/atum/reader/core/PartitioningIdProvider.scala
+++ b/reader/src/main/scala/za/co/absa/atum/reader/core/PartitioningIdProvider.scala
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package za.co.absa.atum.reader.basic
+package za.co.absa.atum.reader.core
 
 import sttp.monad.MonadError
 import sttp.monad.syntax._
@@ -23,7 +23,7 @@ import za.co.absa.atum.model.envelopes.SuccessResponse.SingleSuccessResponse
 import za.co.absa.atum.model.types.basic.AtumPartitions
 import za.co.absa.atum.model.types.basic.AtumPartitionsOps
 import za.co.absa.atum.model.utils.JsonSyntaxExtensions.JsonSerializationSyntax
-import za.co.absa.atum.reader.basic.RequestResult.RequestResult
+import RequestResult.RequestResult
 
 trait PartitioningIdProvider[F[_]] {self: Reader[F] =>
   def partitioningId(partitioning: AtumPartitions)(implicit monad: MonadError[F]): F[RequestResult[Long]] = {
--- a/reader/src/main/scala/za/co/absa/atum/reader/core/RequestResult.scala
+++ b/reader/src/main/scala/za/co/absa/atum/reader/core/RequestResult.scala
@@ -14,25 +14,36 @@
  * limitations under the License.
  */
 
-package za.co.absa.atum.reader.basic
+package za.co.absa.atum.reader.core
 
 import sttp.client3.{DeserializationException, HttpError, Response, ResponseException}
+import sttp.monad.MonadError
 import za.co.absa.atum.model.envelopes.ErrorResponse
+import za.co.absa.atum.reader.exceptions.RequestException.{CirceError, HttpException, ParsingException}
+import za.co.absa.atum.reader.exceptions.RequestException
+import za.co.absa.atum.reader.result.Page
 
 object RequestResult {
-  type CirceError = io.circe.Error
-  type RequestResult[R] = Either[ResponseException[ErrorResponse, CirceError], R]
+  type RequestResult[R] = Either[RequestException, R]
+
+  def RequestOK[T](value: T): RequestResult[T] = Right(value)
+  def RequestFail[T](error: RequestException): RequestResult[T] = Left(error)
 
   implicit class ResponseOps[R](val response: Response[Either[ResponseException[String, CirceError], R]]) extends AnyVal {
     def toRequestResult: RequestResult[R] = {
       response.body.left.map {
         case he: HttpError[String] =>
           ErrorResponse.basedOnStatusCode(he.statusCode.code, he.body) match {
-            case Right(er) => HttpError(er, he.statusCode)
-            case Left(ce)  => DeserializationException(he.body, ce)
+            case Right(er) => HttpException(he.getMessage, he.statusCode, er, response.request.uri)
+            case Left(ce)  => ParsingException.fromCirceError(ce, he.body)
           }
-        case de: DeserializationException[CirceError] => de
+        case de: DeserializationException[CirceError] => ParsingException.fromCirceError(de.error, de.body)
       }
     }
   }
+
+  implicit class RequestPageResultOps[A, F[_]: MonadError](requestResult: RequestResult[Page[A, F]]) {
+    def pageMap[B](f: A => B): RequestResult[Page[B, F]] = requestResult.map(_.map(f))
+  }
+
 }
--- a/reader/src/main/scala/za/co/absa/atum/reader/result/AbstractPage.scala
+++ b/reader/src/main/scala/za/co/absa/atum/reader/result/AbstractPage.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.reader.result
+
+import sttp.monad.MonadError
+
+abstract class AbstractPage [T <: Iterable[_], F[_]: MonadError] {
+    def items: T
+    def hasNext: Boolean
+    def limit: Int
+    def pageStart: Long
+    def pageEnd: Long
+
+    def pageSize: Int = (pageEnd - pageStart).toInt + 1
+    def hasPrior: Boolean = pageStart > 0
+}
+
--- a/reader/src/main/scala/za/co/absa/atum/reader/result/GroupedPage.scala
+++ b/reader/src/main/scala/za/co/absa/atum/reader/result/GroupedPage.scala
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.atum.reader.result
+
+import sttp.monad.MonadError
+import sttp.monad.syntax._
+import za.co.absa.atum.reader.core.RequestResult.{RequestFail, RequestResult}
+import za.co.absa.atum.reader.exceptions.RequestException.NoDataException
+import za.co.absa.atum.reader.result.GroupedPage.GroupPageRoller
+
+import scala.collection.immutable.ListMap
+
+case class GroupedPage[K, V, F[_]: MonadError](
+                                                items: ListMap[K, Vector[V]],
+                                                hasNext: Boolean,
+                                                limit: Int,
+                                                pageStart: Long,
+                                                pageEnd: Long,
+                                                private[reader] val pageRoller: GroupPageRoller[K, V, F]
+                                    ) extends AbstractPage[Map[K, Vector[V]], F] {
+
+  def apply(key: K): Vector[V] = items(key)
+  def keys: Iterable[K] = items.keys
+  def groupCount: Int = items.size
+
+  def map[K1, V1](f: ((K, Vector[V])) => (K1, Vector[V1])): GroupedPage[K1, V1, F] = {
+    val newItems = items.map(f)
+    val newPageRoller: GroupPageRoller[K1, V1, F] = (limit, offset) => pageRoller(limit, offset).map(_.map(_.map(f)))
+    this.copy(items = newItems, pageRoller = newPageRoller)
+  }
+
+  def mapValues[B](f: V => B): GroupedPage[K, B, F] = {
+    def mapper(item: (K, Vector[V])): (K, Vector[B]) = (item._1, item._2.map(f))
+
+    val newItems = items.map(mapper)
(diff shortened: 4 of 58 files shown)


reader/src/main/scala/za/co/absa/atum/reader/core/RequestResult.scala
```diff
@@ -14,25 +14,36 @@
...
+            case Left(ce)  => ParsingException.fromCirceError(ce, he.body)
           }
-        case de: DeserializationException[CirceError] => de
+        case de: DeserializationException[CirceError] => ParsingException.fromCirceError(de.error, de.body)
       }
     }
   }
+
+  implicit class RequestPageResultOps[A, F[_]: MonadError](requestResult: RequestResult[Page[A, F]]) {
+    def pageMap[B](f: A => B): RequestResult[Page[B, F]] = requestResult.map(_.map(f))
```
salamonpavel, line 46: Is this needed given RequestResult is Either therefore map operation is available?

reader/src/main/scala/za/co/absa/atum/reader/result/AbstractPage.scala
```diff
@@ -0,0 +1,31 @@
...
+
+abstract class AbstractPage [T <: Iterable[_], F[_]: MonadError] {
+    def items: T
+    def hasNext: Boolean
+    def limit: Int
+    def pageStart: Long
+    def pageEnd: Long
+
+    def pageSize: Int = (pageEnd - pageStart).toInt + 1
+    def hasPrior: Boolean = pageStart > 0
```
salamonpavel, line 29: What if pageStart is 0 or negative but the absolute value is still lower than pageSize? Could this happen?

reader/src/main/scala/za/co/absa/atum/reader/result/GroupedPage.scala
```diff
@@ -0,0 +1,93 @@
...
+
+import sttp.monad.MonadError
+import sttp.monad.syntax._
+import za.co.absa.atum.reader.core.RequestResult.{RequestFail, RequestResult}
+import za.co.absa.atum.reader.exceptions.RequestException.NoDataException
+import za.co.absa.atum.reader.result.GroupedPage.GroupPageRoller
+
+import scala.collection.immutable.ListMap
+
+case class GroupedPage[K, V, F[_]: MonadError](
```
salamonpavel, line 27: Can you elaborate on the usage of this class? What do we need it for? I often try to limit the scope of my PRs as much as possible.

reader/src/main/scala/za/co/absa/atum/reader/result/Page.scala
```diff
@@ -0,0 +1,96 @@
...
+    if (hasNext) {
+      pageRoller(newPageSize, pageStart + pageSize)
+    } else {
+      MonadError[F].unit(RequestFail(NoDataException("No next page")))
+    }
+  }
+
+  def next: F[RequestResult[Page[T, F]]] = next(limit)
+
+  def +(other: Page[T, F]): Page[T, F] = {
```
salamonpavel, line 66: I think it's not really needed at this point if even in the future ...

reader/src/main/scala/za/co/absa/atum/reader/result/Page.scala
```diff
@@ -0,0 +1,96 @@
...
+case class Page[T, F[_]: MonadError](
+                                      items: Vector[T],
+                                      hasNext: Boolean,
+                                      limit: Int,
+                                      pageStart: Long,
+                                      pageEnd: Long,
+                                      private[reader] val pageRoller: PageRoller[T, F]
+                                    ) extends AbstractPage[Vector[T], F] {
+
+  def apply(index: Int): T = items(index)
```
salamonpavel, line 37: Is it wise to provide apply which might lead to IndexOutOfBoundsExceptions?

reader/src/main/scala/za/co/absa/atum/reader/result/Page.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def map[B](f: T => B): Page[B, F] = {
+    val newItems = items.map(f)
+    val newPageRoller: PageRoller[B, F] = (limit, offset) => pageRoller(limit, offset).map(_.pageMap(f))
+    this.copy(items = newItems, pageRoller = newPageRoller)
+  }
+
+  def prior(newPageSize: Int): F[RequestResult[Page[T, F]]] = {
```
salamonpavel, line 45: I don't think prior is needed. But I might be wrong or maybe I am just not seeing the usage. When I think of Unify I cannot imagine that the backend calls prior given it's stateless.

salamonpavel: As mentioned in Teams I am of the opinion that we should force forward-only consumption pattern to try to avoid repetitive calls to db.

reader/src/main/scala/za/co/absa/atum/reader/result/Page.scala
```diff
@@ -0,0 +1,96 @@
...
+                                      hasNext: Boolean,
+                                      limit: Int,
+                                      pageStart: Long,
+                                      pageEnd: Long,
+                                      private[reader] val pageRoller: PageRoller[T, F]
+                                    ) extends AbstractPage[Vector[T], F] {
+
+  def apply(index: Int): T = items(index)
+
+  def map[B](f: T => B): Page[B, F] = {
```
salamonpavel, line 39: Why do you expect the user to use map on Page if items can be mapped easily anyway?

reader/src/main/scala/za/co/absa/atum/reader/result/Page.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def +(other: Page[T, F]): Page[T, F] = {
+    val newItems = items ++ other.items
+    val newPageStart = pageStart min other.pageStart
+    val newPageEnd = pageEnd max other.pageEnd
+    val newHasNext = hasNext && other.hasNext
+    this.copy(items = newItems, hasNext = newHasNext, pageStart = newPageStart, pageEnd = newPageEnd)
+  }
+
+  def groupBy[K](f: T => K): GroupedPage[K, T, F] = {
```
salamonpavel, line 74: Maybe we could discuss these mapping/grouping stuff first on our tech call?

reader/src/main/scala/za/co/absa/atum/reader/result/Page.scala
```diff
@@ -0,0 +1,96 @@
...
+      limit,
+      pageStart,
+      pageEnd,
+      newPageRoller
+    )
+  }
+}
+
+object Page {
+  type PageRoller[T, F[_]] = (Int, Long) => F[RequestResult[Page[T, F]]]
```
salamonpavel, line 95: Clever.

reader/src/main/scala/za/co/absa/atum/reader/core/PartitioningIdProvider.scala
```diff
@@ -23,7 +23,7 @@ import za.co.absa.atum.model.envelopes.SuccessResponse.SingleSuccessResponse
 import za.co.absa.atum.model.types.basic.AtumPartitions
 import za.co.absa.atum.model.types.basic.AtumPartitionsOps
 import za.co.absa.atum.model.utils.JsonSyntaxExtensions.JsonSerializationSyntax
-import za.co.absa.atum.reader.basic.RequestResult.RequestResult
+import RequestResult.RequestResult
```
lsulak, line 26: I'd prefer this style to be explicit, WDYT?

`import za.co.absa.atum.reader.core.RequestResult.RequestResult`

salamonpavel: I am also in favor of using fully qualified imports.

reader/src/main/scala/za/co/absa/atum/reader/result/Page.scala
```diff
@@ -0,0 +1,96 @@
...
+  def +(other: Page[T, F]): Page[T, F] = {
+    val newItems = items ++ other.items
+    val newPageStart = pageStart min other.pageStart
+    val newPageEnd = pageEnd max other.pageEnd
+    val newHasNext = hasNext && other.hasNext
+    this.copy(items = newItems, hasNext = newHasNext, pageStart = newPageStart, pageEnd = newPageEnd)
+  }
+
+  def groupBy[K](f: T => K): GroupedPage[K, T, F] = {
+    val (newItems, itemsCounts) = items.foldLeft(ListMap.empty[K, Vector[T]], 0) { case ((groupsAcc, count), item) =>
```
lsulak, line 75: Hmm any plans with the `itemsCounts` ? It looks potentially useful but it's nowhere used yet. Alternatively you can just delete and simplify the whole foldLeft block

BTW, if you decide to use and keep it, I'd propose to rename `count` to `countAcc`, just as a common practice (at least mine :D )

reader/src/main/scala/za/co/absa/atum/reader/result/Page.scala
lsulak: I honestly like it and find it very universal

reader/src/test/scala/za/co/absa/atum/reader/core/PartitioningIdProviderUnitTests.scala
```diff
@@ -38,7 +39,7 @@ class PartitioningIdProviderUnitTests extends AnyFunSuiteLike {
   private val atumPartitionsToNotFound = AtumPartitions(List.empty)
 
   private implicit val serverConfig: ServerConfig = ServerConfig(serverUrl)
-  private implicit val monad: IdMonad.type = IdMonad
+  private implicit val monad: MonadError[Identity] = IdMonad
```
lsulak, line 42: very clever-the whole file, just read it for the first time

lsulak: Also in line 71:

`val result: Long = response.getOrElse(throw new Exception("Failed to get partitioning id"))`

I think in tests this is unnecessary, I'd just do something like: `val result: Long = response.get`

lsulak: it's test that simulates something in well-isolated and very specific & deterministic scenario

lsulak: Line 91: isn't this a bit more readable?

`result.left.map(e => assert(e.isInstanceOf[ParsingException]))` instead of `swap`

reader/src/test/scala/za/co/absa/atum/testing/data/PageData.scala
```diff
@@ -0,0 +1,72 @@
...
+  )
+
+  val items2: Vector[TestItem] = Vector(
+    TestItem(3, "aa"),
+    TestItem(4, "bb"),
+    TestItem(5, "dd"),
+    TestItem(4, "cc")
+  )
+
+  def roller(limit: Int, offset: Long): RequestResult[Page[TestItem, Identity]] = {
```
lsulak, line 46: don't you want to make this function more universal? It only works with offset 0 or 6

reader/src/test/scala/za/co/absa/atum/reader/result/AbstractPageUnitTests.scala
```diff
@@ -0,0 +1,53 @@
...
+
+    assert(page.items.size == 3)
+    assert(page.hasNext)
+    assert(page.limit == 3)
+    assert(page.pageStart == 0)
+    assert(page.pageSize == 3)
+    assert(!page.hasPrior)
+
+    val anotherPage = new AbstractPage[Iterable[Int], Identity] {
+      override def items: Iterable[Int] = Seq(1, 2, 3)
```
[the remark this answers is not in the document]
lsulak: I'd propose to implement logically correct tests so that they are understood & the intention for the code under test is well understood - tests are a sort of a documentation as well

reader/src/test/scala/za/co/absa/atum/reader/result/GroupedPageUnitTests.scala
```diff
@@ -0,0 +1,179 @@
+/*
+ * Copyright 2025 ABSA Group Limited
```
lsulak, line 2: 2021

reader/src/test/scala/za/co/absa/atum/testing/implicits/AbstractPageImplicits.scala
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2025 ABSA Group Limited
```
lsulak, line 2: 2021

reader/src/test/scala/za/co/absa/atum/testing/implicits/RequestResultImplicits.scala
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2025 ABSA Group Limited
```
lsulak, line 2: 2021

reader/src/test/scala/za/co/absa/atum/testing/data/PageData.scala
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2025 ABSA Group Limited
```
lsulak, line 2: 2021

reader/src/test/scala/za/co/absa/atum/reader/result/GroupedPageUnitTests.scala
```diff
@@ -0,0 +1,179 @@
...
+    mappedPage1.assertGroupedPage(
+      ListMap(
+        "1" -> Vector("A", "B"),
+        "2" -> Vector("C", "D"),
+        "3" -> Vector("E", "F")
+      ),
+      hasNext = true,
+      limit = 6,
+      pageStart = 0,
+      pageEnd = 5
```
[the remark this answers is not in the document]
lsulak: I guess the overall number of items so that this is correct, just wanted to double check

reader/src/test/scala/za/co/absa/atum/reader/result/GroupedPageUnitTests.scala
```diff
@@ -0,0 +1,179 @@
...
+        2 -> Vector(PageData.TestItem(2, "c"), PageData.TestItem(2, "d")),
+        3 -> Vector(PageData.TestItem(3, "e"), PageData.TestItem(3, "f"))
+      ),
+      hasNext = true,
+      limit = 17,
+      pageStart = 0,
+      pageEnd = 5
+    )
+
+    val Right(priorPage2) = nextPage2.prior(7)
```
lsulak, line 142: Why don't you test scenrio when the limit is for example 4?

reader/src/test/scala/za/co/absa/atum/reader/result/GroupedPageUnitTests.scala
```diff
@@ -0,0 +1,179 @@
...
+    concatenatedPage.assertGroupedPage(
+      ListMap(
+        1 -> Vector(PageData.TestItem(1, "a"), PageData.TestItem(1, "b")),
+        2 -> Vector(PageData.TestItem(2, "c"), PageData.TestItem(2, "d")),
+        3 -> Vector(PageData.TestItem(3, "e"), PageData.TestItem(3, "f"), PageData.TestItem(3, "aa")),
+        4 -> Vector(PageData.TestItem(4, "bb"), PageData.TestItem(4, "cc")),
+        5 -> Vector(PageData.TestItem(5, "dd"))
+      ),
+      hasNext = false,
+      limit = 6,
```
lsulak, line 173: is this logically correct? Number of items is higher than 6

benedeki:
Not considered needed.

PR #340, opened 2025-04-25
monad error as implicit evidence instead of type bound in reader classes

### LINKED ISSUE #339: Move ErrorMonad type bound in Reader classes to evidence based
## Background
The combination of implicit constructor parameters and type bounds on type parameter prevents to explicitly provide the `serverConfig` and `backend` implicit parameters.

## Feature
Move the type bound `MonadError` of the `Reader` classes to evidence based. This should make the work with implicit parameters more flexible.

salamonpavel:
MonadError instance passed as implicit evidence instead of type bound.
Closes #339 

Release notes:
- Reader classes' signatures changed to receive MonadError instance as implicit evidence instead of type bound


benedeki (approved):
Thank you. Love that you bothered to even add a test case.

PR #341, opened 2025-04-25
migration script for measure definitions

### LINKED ISSUE #336: Measure names are not saved properly to DB
## Describe the bug
When a a measure is added to `AtumContext` it's not saved properly on to DB.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Create an `AtumContext`
2. Add a measure `count`
3. Create a checkpoint using the context
4. Check the DB to see the measure is saved as `*`

## Expected behavior
The `count` function is saved as measure name.

## Additional fix
Include a SQL migration script that will replace all '*" with `count`.

salamonpavel:
Prepares migration script that fixes measure names.
Closes #336 

Release notes:
- Introduces migration script that fixes incorrectly stored measure names in DB.


salamonpavel:
Will be done in Unify

PR #344, opened 2025-04-28
measures as normal classes

### LINKED ISSUE #342: Measures can be created with custom measureName
## Describe the bug
`Measures` can be created with custom `measureName` which saves the broken data causing eventual problems 

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Create an `AtumContext`
2. Add a measure `RecordCount('*')` to it
3. Create checkpoint
4. Check data in DB
5. See error - measure count was save under the name *
 
## Expected behavior
It's not possible to create a defined measure using a custom name.

## Reason

> _Scala 2.12 does not support private constructors for case classes directly. This is because case classes in Scala 2.12 automatically generate a public apply method in their companion object, which allows instantiation even if the constructor is marked as private._

### Use normal classes instead of case classes
```sql
final class RecordCount private (val measureName: String) extends AtumMeasure {
    private val columnExpression = count("*")

    override def function: MeasurementFunction =
      (ds: DataFrame) => {
        val resultValue = ds.select(columnExpression).collect()
        MeasureResult(resultValue(0).toString, resultValueType)
      }

    override def measuredColumns: Seq[String] = Seq.empty
    override val resultValueType: ResultValueType = ResultValueType.LongValue
  }
object RecordCount {
    private[agent] val measureName: String = "count"
    def apply(): RecordCount = new RecordCount(measureName)
}
```

### Create a private `apply` method similar to the constructor
```scala
object RecordCount {
  private[agent] val measureName: String = "count"
  private def apply(measureName: String): RecordCount = new RecordCount(measureName)
  def apply(): RecordCount = new RecordCount(measureName)
}
```

## Suggestion
Make `RecordCount` and object instead of (case) class.

**NB!**
Verify that SerDe works after the change.

salamonpavel:
Changes case classes to normal class for measure classes.
Closes #342 

Release notes:
- Due to missing support for private constructors for case classes in Scala 2.12 measure classes were refactored from case classes to standard classes in order to avoid having their private constructor accessible.

### DIFF
--- a/agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
+++ b/agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
@@ -39,7 +39,8 @@ final case class UnknownMeasure(measureName: String, measuredColumns: Seq[String
 
 object AtumMeasure {
 
-  case class RecordCount private (measureName: String) extends AtumMeasure {
+  final case object RecordCount extends AtumMeasure {
+    val measureName: String = "count"
     private val columnExpression = count("*")
 
     override def function: MeasurementFunction =
@@ -51,12 +52,8 @@ object AtumMeasure {
     override def measuredColumns: Seq[String] = Seq.empty
     override val resultValueType: ResultValueType = ResultValueType.LongValue
   }
-  object RecordCount {
-    private[agent] val measureName: String = "count"
-    def apply(): RecordCount = RecordCount(measureName)
-  }
 
-  case class DistinctRecordCount private (measureName: String, measuredCols: Seq[String]) extends AtumMeasure {
+  final class DistinctRecordCount private (val measureName: String, measuredCols: Seq[String]) extends AtumMeasure {
     require(measuredCols.nonEmpty, "At least one measured column has to be defined.")
 
     private val columnExpression = countDistinct(col(measuredCols.head), measuredCols.tail.map(col): _*)
@@ -72,10 +69,10 @@ object AtumMeasure {
   }
   object DistinctRecordCount {
     private[agent] val measureName: String = "distinctCount"
-    def apply(measuredCols: Seq[String]): DistinctRecordCount = DistinctRecordCount(measureName, measuredCols)
+    def apply(measuredCols: Seq[String]): DistinctRecordCount = new DistinctRecordCount(measureName, measuredCols)
   }
 
-  case class SumOfValuesOfColumn private (measureName: String, measuredCol: String) extends AtumMeasure {
+  final class SumOfValuesOfColumn private (val measureName: String, measuredCol: String) extends AtumMeasure {
     private val columnAggFn: Column => Column = column => sum(column)
 
     override def function: MeasurementFunction = (ds: DataFrame) => {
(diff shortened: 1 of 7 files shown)


agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -69,13 +66,25 @@ object AtumMeasure {
...
+      case that: DistinctRecordCount =>
+        measureName == that.measureName &&
+        measuredCols == that.measuredCols
+      case _ => false
+    }
+
+    override def hashCode(): Int = {
+      val state = Seq(measureName, measuredCols)
+      state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
+    }
```
benedeki, line 80: Couldn't these two things be part of `AtumMeasure`?

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -35,11 +35,12 @@ sealed trait Measure {
...
 
 final case class UnknownMeasure(measureName: String, measuredColumns: Seq[String], resultValueType: ResultValueType)
-  extends Measure
+    extends Measure
 
 object AtumMeasure {
 
-  case class RecordCount private (measureName: String) extends AtumMeasure {
+  final case object RecordCount extends AtumMeasure {
+    val measureName: String = "count"
```
benedeki, line 43: To make it look like the other functions:
```suggestion
    val measureName: String = "count"
    
    def apply() = this
```
Maybe with some extra tweaking like declaring `final class RecordCount` and making final case object RecordCount extends RecordCount`.

benedeki (commented):
Please make the Release notes readable as release note - desciription what has changed and why 😉

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
benedeki: Also makes the change backward compatible wherever `RecordCount()` was used.

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
salamonpavel: good idea

salamonpavel:
> Please make the Release notes readable as release note - desciription what has changed and why 😉

hope it's better now :)

agent/src/main/scala/za/co/absa/atum/agent/model/Measure.scala
```diff
@@ -51,6 +63,12 @@ object AtumMeasure {
 
     override def measuredColumns: Seq[String] = Seq.empty
     override val resultValueType: ResultValueType = ResultValueType.LongValue
+
+    override protected def productIterator: Iterator[Any] = Iterator(measureName)
+  }
+  object RecordCount {
+    private[agent] val measureName: String = "count"
+    def apply(): RecordCount = new RecordCount(measureName)
```
benedeki, line 71: What I meant:
```suggestion
    override protected def productIterator: Iterator[Any] = Iterator(measureName)
    
    def apply(): RecordCount = this
  }
  object RecordCount extends RecordCount("count") {
```

benedeki (approved):

PR #385, opened 2025-10-15
additional data endpoints response structure

### LINKED ISSUE #365: Redundant `data` sub-structure in `additional-data` APIs
## Describe the bug

The response of the following two API endpoints return redundant `data -> data` structure:
* PATCH `/api/v2/partitionings/{{partitioning_id}}/additional-data`
* GET `/api/v2/partitionings/{{partitioning_id}/additional-data`

We probably want only 1 `data` level.

## To Reproduce
Just call any of these two API endpoints and check the response.

## Expected behavior
Only 1 nesting level.

## Screenshots

The response is as follows:

```
{
    "data": {
        "data": {
            "additionalData1": {
                "value": "firstItem",
                "author": "apiTest1ShotAD"
            },
            "additionalData2": {
                "value": "secondItem",
                "author": "apiTest1ShotAD"
            },
            "additionalData3": {
                "value": "thirdItem",
                "author": "apiTest1ShotAD"
            }
        }
    },
    "requestId": "a11cb0dc-f48b-4189-9078-44b3ee214b19"
}
```

salamonpavel:
Release notes:
- Fixes response structure in v2 get/patch additional data endpoints

Fixes #365 

NOTE: I am not sure it's a bug and I think it should be kept as it is. So let's discuss...

### DIFF
--- a/server/src/main/scala/za/co/absa/atum/server/api/v2/http/Endpoints.scala
+++ b/server/src/main/scala/za/co/absa/atum/server/api/v2/http/Endpoints.scala
@@ -70,7 +70,7 @@ object Endpoints extends BaseEndpoints {
 
   val patchPartitioningAdditionalDataEndpoint
     : PublicEndpoint[(Long, AdditionalDataPatchDTO), ErrorResponse, SingleSuccessResponse[
-      Map[String,Option[AdditionalDataItemDTO]]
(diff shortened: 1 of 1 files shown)


ABLL526 (commented):
I am happy with these changes.
As long as the unit tests passed with this altered DTO, then I will approve.

PR #386, opened 2025-10-17
optional endpoint params

### LINKED ISSUE #364: Optional query parameters for pagination are actually mandatory
## Describe the bug
These endpoints don't work without limit and offset query parameters! They should be optional, but are not. So I had to add `?limit=10&offset=0` into the queries for them to work:
* GETs `/api/v2/partitionings/{{partitioning_id}}/checkpoints` 
* GETs `/api/v2/flows/{{flow_id}}/partitionings`
* GETs `/api/v2/flows/{{flow_id}}/checkpoints`

## To Reproduce
Just call any of these 3 API endpoints without one / both of these parameters and you will see that they don't work.

## Expected behavior
Sensible defaults are used when not specifying these parameters, i.e. the endpoints work even without them.

salamonpavel:
Fixes optional limit and offset params for several endpoints. Tapir's `.default` works with non-optional types only
Fixes #364

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @benedeki @lsulak @Zejnilovic @dk1844 @salamonpavel @ABLL526
+* @benedeki @lsulak @Zejnilovic @dk1844 @salamonpavel @ABLL526 @MatloaItumeleng
--- a/api-tests/1_shot/Atum-Service_1-Shot_test_flow.postman_collection.json
+++ b/api-tests/1_shot/Atum-Service_1-Shot_test_flow.postman_collection.json
@@ -1449,7 +1449,7 @@
 						"header": [],
 						"body": {
 							"mode": "raw",
-							"raw": "{\n  \"parentPartitioningId\": \"{{partitioning_id}}\",\n  \"author\": \"{{ad_author}}\",\n  \"copyMeasurements\": true,\n  \"copyAdditionalData\": true\n}",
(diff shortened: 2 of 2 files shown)


server/src/main/scala/za/co/absa/atum/server/api/v2/http/Endpoints.scala
```diff
@@ -101,27 +101,28 @@ object Endpoints extends BaseEndpoints {
...
   val getPartitioningCheckpointsEndpoint
-    : PublicEndpoint[(Long, Option[Int], Option[Long], Option[String]), ErrorResponse, PaginatedResponse[
+    : PublicEndpoint[(Long, Int, Long, Option[String]), ErrorResponse, PaginatedResponse[
       CheckpointV2DTO
     ], Any] = {
     apiV2.get
       .in(V2Paths.Partitionings / path[Long]("partitioningId") / V2Paths.Checkpoints)
-      .in(query[Option[Int]]("limit").default(Some(10)).validateOption(Validator.inRange(1, 1000)))
-      .in(query[Option[Long]]("offset").default(Some(0L)).validateOption(Validator.min(0L)))
+      // Tapir's .default works with non-optional types only
```
lsulak: Thanks for the comment, ok I liked when it was explicitly said in the type (= `Option`), but if this is how Tapir works, then it is the way forward. Thanks for this!

MatloaItumeleng (approved):
LGTM

Issue #343, opened 2025-04-26
Shading is needed when integrating Agent with Spark 3.x versions


benedeki:
## Describe the bug
Dependency shading might be needed when using the Agent in Spark environment (especially when some Hadoop dependencies are in use as well, for example if you package the application that contains the Agent and use such JAR in Spark Submit command). 

## Expected behavior
Shading is not needed. The required libraries that needs shading are part of the Agent library already in shaded form.

## Workaround (shading to add to the project)
Here's the suggested project code snippet for Maven:

```xml
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-shade-plugin</artifactId>
                <version>${maven.shade.plugin.version}</version>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>shade</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <relocations>
                        <relocation>
                            <pattern>okhttp3</pattern>
                            <shadedPattern>shaded.okhttp3</shadedPattern>
                        </relocation>
                        <relocation>
                            <pattern>okio</pattern>
                            <shadedPattern>shaded.okio</shadedPattern>
                        </relocation>
                        <relocation>
                            <pattern>sttp</pattern>
                            <shadedPattern>shaded.sttp</shadedPattern>
                        </relocation>
                        <relocation>
                            <pattern>cats</pattern>
                            <shadedPattern>shaded.cats</shadedPattern>
                        </relocation>
                        <relocation>
                            <pattern>shapeless</pattern>
                            <shadedPattern>shaded.shapeless</shadedPattern>
                        </relocation>
                        <relocation>
                            <pattern>kotlin</pattern>
                            <shadedPattern>shaded.kotlin</shadedPattern>
                        </relocation>
                    </relocations>
                    <filters>
                        <filter>
                            <artifact>*:*</artifact>
                            <excludes>
                                <exclude>META-INF/*.SF</exclude>
                                <exclude>META-INF/*.DSA</exclude>
                                <exclude>META-INF/*.RSA</exclude>
                            </excludes>
                        </filter>
                    </filters>
                </configuration>
            </plugin>
```

PR #388, opened 2025-10-20
unify versions to eliminate shading needs

salamonpavel:
In order to eliminate needs for shading when using Atum Agent dependencies used also in Unify have been upgraded to versions used in Unify. When released we can try to remove the shading.

Release notes:
- Upgrades several dependencies to higher versions

~Closes~ Relates to #343


benedeki (approved):
I think the increase in versions is a good thing. I am just not sure it solves the original issue. I replaced the _Closes_ with _Relates to_.

PR #406, opened 2026-01-05
create partitioning (atum context) using v2 endpoints

### LINKED ISSUE #347: Switch `atumContext` spawning to use API v2
## Background
A clear and concise description of where the limitation lies.

## Feature
A description of the requested feature.

## Example [Optional]
A simple example if applicable.

Solution Ideas:
1. 
2. 
3.

salamonpavel:
Release notes:
- `createPartitioning` operation in `HttpDispatcher` uses v2 endpoints

Closes #347

### DIFF
--- a/.github/workflows/jacoco_report.yml
+++ b/.github/workflows/jacoco_report.yml
@@ -87,7 +87,7 @@ jobs:
       postgres:
         image: postgres:15
         env:
-          POSTGRES_PASSWORD: postgres
+          POSTGRES_PASSWORD: changeme
           POSTGRES_DB: atum_db
         options: >-
           --health-cmd pg_isready


agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
```diff
@@ -37,45 +36,99 @@ class HttpDispatcher(config: Config) extends Dispatcher(config) with Logging {
...
   private def createAdditionalDataEndpoint(partitioningId: Long): Uri =
-    Uri.unsafeParse(s"$serverUrl$apiV2/${ApiPaths.V2Paths.Partitionings}/$partitioningId/${ApiPaths.V2Paths.AdditionalData}")
+    Uri.unsafeParse(
+      s"$serverUrl$apiV2/${ApiPaths.V2Paths.Partitionings}/$partitioningId/${ApiPaths.V2Paths.AdditionalData}"
+    )
+  private val postPartitioningEndpoint = Uri.unsafeParse(s"$serverUrl$apiV2/${ApiPaths.V2Paths.Partitionings}")
+  private def createGetPartitioningMeasuresEndpoint(partitioningId: Long) = Uri.unsafeParse(
+    s"$serverUrl$apiV2/${ApiPaths.V2Paths.Partitionings}/$partitioningId/${ApiPaths.V2Paths.Measures}"
+  )
+  private def createGetPartitioningAdditionalDataEndpoint(partitioningId: Long) = Uri.unsafeParse(
```
lsulak: this is exactly the same as one above, `createAdditionalDataEndpoint`

actually, what's the value in these? It's to me really unnecessary and most of these are only used on 1 place, so not reusable. And they are only for this module. I would say this is added tiny bit of a complexity for no real benefit (I know that it was here before, but I revisited it just now)

What do you think?

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
```diff
@@ -96,7 +149,7 @@ class HttpDispatcher(config: Config) extends Dispatcher(config) with Logging {
     log.debug(s"Got partitioning ID: '$partitioningId'")
 
     val request = commonAtumRequest
-      .patch(createAdditionalDataEndpoint(partitioningId))
+      .patch(createAdditionalDataEndpoint(partitioningId.get))
```
lsulak: this can fail - if we want to update AD of non-existing partitioning, the code will fail on `.get` operation. This ideally should be implemented in different way

agent/src/test/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcherUnitTests.scala
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  def dispatcher: HttpDispatcher = new HttpDispatcher(mockConfig) {
+    override private[dispatcher] val backend = mockBackend
+  }
+
+  def isGetPartitioningRequest(
+    partitioning: Seq[PartitionDTO]
+  ): Request[Either[String, String], capabilities.WebSockets] =
+    argThat(new org.mockito.ArgumentMatcher[Request[Either[String, String], sttp.capabilities.WebSockets]] {
```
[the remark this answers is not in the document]
lsulak: but I don't insist, it all looks nice and clean, also checked the other similar methods

lsulak (commented):
I finished the review, I haven't found anything major. Besides these changes, I think this is complete (will reiterate for the rest of the team):
* Integration tests / compatibility between Agent <-> Server, should not be affected. The interfaces didn't change, only the internals.
* Thus also 1-shot API tests shouldn't be affected because we didn't change any API, just the internals of agent.

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
salamonpavel: addressed

agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
salamonpavel: obsolete

PR #407, opened 2026-01-06
Feature/348 create checkpoint v2

### LINKED ISSUE #348: Switch `createCheckpoint` to use API v2
## Background
A clear and concise description of where the limitation lies.

## Feature
A description of the requested feature.

## Example [Optional]
A simple example if applicable.

Solution Ideas:
1. 
2. 
3.

salamonpavel:
Release notes:
- `saveCheckpoint` operation in `HttpDispatcher` uses v2 endpoints

Closes #348

### DIFF
--- a/agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
+++ b/agent/src/main/scala/za/co/absa/atum/agent/AtumContext.scala
@@ -164,7 +164,7 @@ class AtumContext private[agent] (
     val retrievedAD = agent.updateAdditionalData(this.atumPartitions, currAdditionalDataSubmit)
 
     // Could be different from the one that was submitted. Replacing, just to have the most fresh copy possible.
-    this.additionalData = retrievedAD.data.map { case (k, v) => (k, v.map(_.value)) }.toMap
+    this.additionalData = retrievedAD.data.map { case (k, v) => (k, v.map(_.value)) }
     this
   }
 


agent/src/main/scala/za/co/absa/atum/agent/dispatcher/ConsoleDispatcher.scala
```diff
@@ -18,15 +18,7 @@ package za.co.absa.atum.agent.dispatcher
...
-import za.co.absa.atum.model.dto.{
-  AdditionalDataDTO,
-  AdditionalDataItemDTO,
-  AdditionalDataPatchDTO,
-  AtumContextDTO,
-  CheckpointDTO,
-  PartitioningDTO,
-  PartitioningSubmitDTO
-}
+import za.co.absa.atum.model.dto._
```
lsulak: Do you prefer this style in general?

project/FlywayConfiguration.scala
```diff
@@ -21,7 +21,7 @@ object FlywayConfiguration {
   private val database = "atum_db"
   val flywayUrl = s"jdbc:postgresql://$host:$port/$database"
   val flywayUser = "postgres"
-  val flywayPassword = "postgres"
+  val flywayPassword = "changeme"
```
lsulak: I see that you made this change on all places, what's the reason?

server/src/main/scala/za/co/absa/atum/server/api/v2/http/Endpoints.scala
```diff
@@ -60,23 +60,21 @@ object Endpoints extends BaseEndpoints {
   }
 
   val getPartitioningAdditionalDataEndpoint
-    : PublicEndpoint[Long, ErrorResponse, SingleSuccessResponse[AdditionalDataDTO.Data], Any] = {
+    : PublicEndpoint[Long, ErrorResponse, MultiSuccessResponse[AdditionalDataItemV2DTO], Any] = {
     apiV2.get
       .in(V2Paths.Partitionings / path[Long]("partitioningId") / V2Paths.AdditionalData)
       .out(statusCode(StatusCode.Ok))
-      .out(jsonBody[SingleSuccessResponse[AdditionalDataDTO.Data]])
+      .out(jsonBody[MultiSuccessResponse[AdditionalDataItemV2DTO]])
```
lsulak: On one hand, I liked the `SingleSuccessResponse[AdditionalDataDTO.Data]` because it 'restricts' the duplication of same AD key, by the 'contract' it's immediatelly & explicitly obvious. 

On another hand, I like `MultiSuccessResponse[AdditionalDataItemV2DTO]]` because it is more 'explicit' and precise on what is key, author, and value - and can be extended easily. Map of X and Y cannot be as easily. 

So let's carry on with the implementation. It's same as for Measures. But it's important to know that we changed the contract here.

model/src/main/scala/za/co/absa/atum/model/dto/AdditionalDataItemV2DTO.scala
```diff
@@ -0,0 +1,15 @@
+package za.co.absa.atum.model.dto
+
+import io.circe.{Decoder, Encoder}
+import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
+
+case class AdditionalDataItemV2DTO(
```
lsulak: Ok, this structure makes sense to me.

But when I look at `model/` now, it's very ... rich. I think that `AdditionalDataDTO` is only used for AD Update. What do you think about renaming it to `UpdateAdditionalDataDTO` ? 

Please double check.

agent/src/test/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcherUnitTests.scala
```diff
@@ -2,17 +2,21 @@ package za.co.absa.atum.agent.dispatcher
...
 import com.typesafe.config.Config
 import sttp.capabilities
+import za.co.absa.atum.agent.exception.AtumAgentException.HttpException
 import za.co.absa.atum.model.dto._
 import za.co.absa.atum.model.envelopes.SuccessResponse.{MultiSuccessResponse, SingleSuccessResponse}
 import za.co.absa.atum.model.utils.JsonSyntaxExtensions.JsonSerializationSyntax
 
+import java.time.ZonedDateTime
+
+@Ignore // has to be run with a real server
```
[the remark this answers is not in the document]
salamonpavel: it's a newly introduced test class and actually everything is mocked

PR #439, opened 2026-05-12
Added factory for config-backed Atum-Agent

ABLL526:
## Task:
Adds a factory method to create an `AtumAgent` instance backed by an explicitly provided `Typesafe Config`, enabling multiple agents with different dispatcher setups within the same JVM.

## Context:
- https://github.com/absa-group/ursa-unify-deployment/pull/400
- https://github.com/absa-group/AUL/pull/4930

The above PRs updates the deployment pipeline to extract, package, and publish the external TypeSafe config for the data-measurement Spark plugin and ensure it is shipped to Spark via `--files`:

This supports the “external first, classpath fallback” strategy within AUL: the plugin first attempts to load `application.conf` distributed through Spark `--files`, and falls back to the classpath config if no external file is available.

Essentially, we want to externalize all configuration files. So, it is easier for the Ops teams and Dev teams to make changes to the configuration files without a need to always release a new version of Atum-Services or Unify. But since Atum-Services is in the codebase of Unify, a small change is needed on Atum-Services to be able to externalize these config files.

## Closes: 
#438 

## Release Notes:
- Added factory for config-backed agent
  -  Added function `fromConfig` to `AtumAgent`
  -  Added Unit Tests
  -  Updated `README` to include Option 3

### DIFF
--- a/agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
+++ b/agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
@@ -131,4 +131,7 @@ object AtumAgent extends AtumAgent {
     }
   }
 
+  def fromConfig(config: Config): AtumAgent = new AtumAgent {
+    override val dispatcher: Dispatcher = dispatcherFromConfig(config)
+  }
 }
--- a/agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
+++ b/agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
@@ -66,7 +66,7 @@ trait AtumAgent {
    *  Provides an AtumContext given a `AtumPartitions` instance. Retrieves the data from AtumService API.
    *
    *  Note: if partitioning doesn't exist in the store yet, a new one will be created with the author stored in
-   *    `AtumAgent.currentUser`. If partitioning already exists, this attribute will be ignored because there
+   *    `currentUser`. If partitioning already exists, this attribute will be ignored because there
    *    already is an author who previously created the partitioning in the data store. Each Atum Context thus
    *    can have different author potentially.
    *
@@ -75,7 +75,7 @@ trait AtumAgent {
    *          partitioning already existed.
    */
   def getOrCreateAtumContext(atumPartitions: AtumPartitions): AtumContext = {
-    val authorIfNew = AtumAgent.currentUser
+    val authorIfNew = this.currentUser
     val partitioningDTO = PartitioningSubmitDTO(atumPartitions.toPartitioningDTO, None, authorIfNew)
 
     val atumContextDTO = dispatcher.createPartitioning(partitioningDTO)
@@ -92,7 +92,7 @@ trait AtumAgent {
    *  @return Atum context object
    */
   def getOrCreateAtumSubContext(subPartitions: AtumPartitions)(implicit parentAtumContext: AtumContext): AtumContext = {
-    val authorIfNew = AtumAgent.currentUser
+    val authorIfNew = this.currentUser
     val newPartitions: AtumPartitions = parentAtumContext.atumPartitions ++ subPartitions
 
     val newPartitionsDTO = newPartitions.toPartitioningDTO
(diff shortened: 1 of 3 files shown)


agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -120,7 +120,7 @@ trait AtumAgent {
 
 object AtumAgent extends AtumAgent {
 
-  override val dispatcher: Dispatcher = dispatcherFromConfig()
+  override lazy val dispatcher: Dispatcher = dispatcherFromConfig()
 
   def dispatcherFromConfig(config: Config = ConfigFactory.load()): Dispatcher = {
```
lsulak: I checked, no one is using this method. I don't actually find it useful when I think about it that much. It is actually quite misleading in my opinion (or was, before your PR). 

Users might call this and retrieve the dispatcher, and then what? They can still work with the original singleton Agent and they cannot inject it into it so they cannot do anythinh useful with it. But I LIKE your approach and `fromConfig` method. Maybe keep that method `dispatcherFromConfig` and use it like you do, but change it to be private: `private[agent] def dispatcherFromConfig` to be super explicit about the intended use

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -120,7 +120,7 @@ trait AtumAgent {
 
 object AtumAgent extends AtumAgent {
 
-  override val dispatcher: Dispatcher = dispatcherFromConfig()
+  override lazy val dispatcher: Dispatcher = dispatcherFromConfig()
```
lsulak: I don't think this matters. Singleton always uses default config — no point delaying it. And your `fromConfig` initializes and works with different Agent instance. So let's be honest & clear about this `val` - I cannot come up with any reason why it should/could be `lazy`. What do you think? / did think about ?

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -131,4 +131,7 @@ object AtumAgent extends AtumAgent {
     }
   }
 
+  def fromConfig(config: Config): AtumAgent = new AtumAgent {
+    override lazy val dispatcher: Dispatcher = dispatcherFromConfig(config)
```
lsulak: `lazy` here is also unnecessary in my view. Since the user explicitly called `fromConfig` with a config he/she supplied, they clearly intend to use it — delaying initialization gains nothing.

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
```diff
@@ -131,4 +131,7 @@ object AtumAgent extends AtumAgent {
...
+  }
```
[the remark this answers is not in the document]
lsulak: This is actually super true, @ABLL526. SOme pieces of codebase work with the singleton, but in `fromConfig` you are creating a new instance of Agent... so, what Copilot identified is a serious concern that needs to be addressed.

Find these `AtumAgent.` and correct this please.

lsulak (commented):
I like this, thank you for your contribution!!

One thing, since you fundamentally changed this: before we had only 1 singleton Atum Agent, now a user can create his own instance. This is not covered with tests, so there is no verification that the returned agent is independent (own context store, own currentUser resolution). Worth adding at least a test showing two agents with different configs coexist.

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
ABLL526: I agree, the only reason was to ensure the PR worked on its own. I am changing it so as to not delay it.

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
ABLL526: Agreed, changed it to private as suggested

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
ABLL526: Agreed as above, made the change

agent/src/main/scala/za/co/absa/atum/agent/AtumAgent.scala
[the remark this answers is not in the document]
ABLL526: Understood, made the changes I hope I covered it all.

lsulak (approved):
Approving, thanks! With your prior permissions (on a chat), I'm gonna merge this now.

PR #443, opened 2026-05-22
#205: adding retry mechanism for Atum agent on HTTP dispatcher level

### LINKED ISSUE #205: Release process is broken
## Describe the bug
The release action did not complete to get out Atum 0.2

## To Reproduce
Hard to reproduce.
See log: https://github.com/AbsaOSS/atum-service/actions/runs/9363398772/job/25774102738#step:4:194

## Expected behavior
Release succeeds.

lsulak:
Closes: #205

# Release Notes:
* Atum Service: adding DB connection timeout parameter for being able to tune this on the server side
* Atum Agent: adding a simple retry mechanism onto the agent's HTTP layer and also changeable settings for timeouts on the HTTP layer (for the underlying OkHTTP client)

### DIFF
--- a/agent/src/main/resources/reference.conf
+++ b/agent/src/main/resources/reference.conf
@@ -20,3 +20,12 @@ atum.dispatcher.type="http"
 # Maximum number of dispatch captures to keep in memory
 #atum.dispatcher.capture.capture-limit=1000 # 0 means no limit
 
+# Retry configuration for the HTTP dispatcher (all keys optional — defaults apply if absent)
+#atum.dispatcher.http.retry.max-retries=3          # number of retries after first failure (0 = no retries)
+#atum.dispatcher.http.retry.initial-delay-ms=1000  # initial backoff delay in milliseconds
+#atum.dispatcher.http.retry.max-delay-ms=10000     # maximum backoff delay cap in milliseconds
+
+# HTTP timeout configuration (all keys optional — defaults apply if absent)
+#atum.dispatcher.http.timeout.connect-ms=10000     # TCP connection timeout in milliseconds (default: 10 000)
+#atum.dispatcher.http.timeout.read-ms=30000        # response read timeout in milliseconds (default: 30 000)
+#atum.dispatcher.http.timeout.write-ms=10000       # request write timeout in milliseconds (default: 10 000)
--- a/agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
+++ b/agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
@@ -153,7 +153,11 @@ class HttpDispatcher(config: Config) extends Dispatcher(config) with Logging {
       .body(checkpointV2DTO.asJsonString)
 
     val response = withRetry(request)
-    handleResponseBody(response)
+    // 409 Conflict means the checkpoint was already saved (e.g. prior attempt succeeded but response was lost).
+    // This is expected on retry and should not be treated as an error.
+    if (response.code != StatusCode.Conflict) {
+      handleResponseBody(response)
+    }
   }
 
   override protected[agent] def updateAdditionalData(
--- a/agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
+++ b/agent/src/main/scala/za/co/absa/atum/agent/dispatcher/HttpDispatcher.scala
@@ -17,6 +17,7 @@
 package za.co.absa.atum.agent.dispatcher
 
 import com.typesafe.config.Config
+import okhttp3.OkHttpClient
 import org.apache.spark.internal.Logging
 import sttp.capabilities
 import sttp.client3._
@@ -28,6 +29,8 @@ import za.co.absa.atum.model.ApiPaths
 import za.co.absa.atum.model.dto._
 import za.co.absa.atum.model.envelopes.SuccessResponse.{MultiSuccessResponse, SingleSuccessResponse}
 import za.co.absa.atum.model.utils.JsonSyntaxExtensions._
+
+import java.util.concurrent.TimeUnit
 
 class HttpDispatcher(config: Config) extends Dispatcher(config) with Logging {
   import HttpDispatcher._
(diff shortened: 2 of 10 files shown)


agent/src/main/resources/reference.conf
```diff
@@ -20,3 +20,12 @@ atum.dispatcher.type="http"
...
 
+# Retry configuration for the HTTP dispatcher (all keys optional — defaults apply if absent)
+#atum.dispatcher.http.retry.max-retries=3          # number of retries after first failure (0 = no retries)
+#atum.dispatcher.http.retry.initial-delay-ms=1000  # initial backoff delay in milliseconds
+#atum.dispatcher.http.retry.max-delay-ms=10000     # maximum backoff delay cap in milliseconds
+
+# HTTP timeout configuration (all keys optional — defaults apply if absent)
+#atum.dispatcher.http.timeout.connect-ms=10000     # TCP connection timeout in milliseconds (default: 10 000)
+#atum.dispatcher.http.timeout.read-ms=30000        # response read timeout in milliseconds (default: 30 000)
+#atum.dispatcher.http.timeout.write-ms=10000       # request write timeout in milliseconds (default: 10 000)
```
benedeki, line 31: Sorry I am not sure, what exactly mean read and write timeout?

agent/src/main/resources/reference.conf
```diff
@@ -20,3 +20,12 @@ atum.dispatcher.type="http"
 # Maximum number of dispatch captures to keep in memory
 #atum.dispatcher.capture.capture-limit=1000 # 0 means no limit
 
+# Retry configuration for the HTTP dispatcher (all keys optional — defaults apply if absent)
+#atum.dispatcher.http.retry.max-retries=3          # number of retries after first failure (0 = no retries)
```
benedeki, line 24: My personal option would be, to have default on 1 - generally in a healthy, stable well load balanced environment, there should not be need for more - but not a hard complain.

salamonpavel (commented):
# PR #443 Review — Retry Mechanism & DB Connection Timeout

## Summary

This PR introduces two welcome improvements:

1. **Agent-side HTTP retry with exponential backoff** — transient 5xx responses and `IOException`s (connection resets, read timeouts) are now retried up to 3 times with jitter, instead of failing immediately.
2. **Configurable OkHTTP timeouts & HikariCP `connectionTimeout`** — operators can now tune connect/read/write timeouts on the agent and connection-acquire timeout on the server without code changes.

Both changes are well-implemented: config keys are optional with sensible defaults, the 409-idempotency handling for `saveCheckpoint` is a nice touch, and the test coverage is solid.

---

## Why client-side retries alone are not enough

While retries improve resilience against **transient** failures (brief network blips, single-request timeouts), they can **amplify** sustained overload:

| Scenario | Without retries | With retries (×3) |
|---|---|---|
| 100 agents hit a struggling server | 100 requests | Up to **400 requests** |
| Server returns 503 under load | Agents back off (by failing) | Agents retry 3× — **more** load on an already-overloaded server |
| DB connection pool exhausted | Requests fail at HikariCP timeout | Retried requests queue up again — pool stays saturated longer |

The fundamental issue: **the server currently has zero admission control**. It accepts unlimited concurrent requests, all of which compete for 10 HikariCP connections. Client-side retries help survive momentary blips but worsen sustained pressure.

Lowering `connectionTimeout` from 30s → 10s is helpful (fail-fast), but it only changes *when* the failure happens — it doesn't prevent the overload itself.

---

## Recommended server-side complement

The most effective protection is a **concurrency semaphore at the transactor level** — a single gate that limits how many queries can be in-flight across the entire application, matching the DB connection pool size.

### Why the transactor level?

All database calls in the server flow through a single `HikariTransactor[Task]` provided by `TransactorProvider`. Wrapping it with a ZIO [`Semaphore`](https://zio.dev/reference/concurrency/semaphore/) means:

- **One change, all services protected** — no need to modify every controller or service class.
- **Cumulative limit** — all endpoints share the same permit pool.
- **Clean backpressure** — excess fibers suspend on the semaphore instead of queuing inside HikariCP, giving you the option to fail fast with a timeout.

### How it works

```scala
// In TransactorProvider.layer — after creating the HikariTransactor:
import zio.Semaphore
import cats.~>

for {
  // ... existing setup ...
  xa  <- HikariTransactor.fromHikariConfig[Task](hikariConfig, ...).toScopedZIO

  // Wrap: at most maxPoolSize queries run concurrently
  sem <- Semaphore.make(postgresConfig.maxPoolSize)
  throttledXa = xa.mapK(new ~>[Task, Task] {
    def apply[A](fa: Task[A]): Task[A] = sem.withPermit(fa)
  })
} yield throttledXa
```

### Optional: fail-fast under sustained load

Instead of queueing indefinitely, reject with a clear error when the server is saturated:

```scala
sem.withPermit(fa)
   .timeoutFail(GeneralDatabaseError("server overloaded — try again later"))(5.seconds)
```

This turns into an HTTP 503 at the controller level, which the agent's new retry mechanism will handle gracefully — creating a healthy feedback loop:

```
Server overloaded → 503 → Agent backs off with jitter → Server recovers → Next retry succeeds
```

### Relevant ZIO documentation

- [`Semaphore`](https://zio.dev/reference/concurrency/semaphore/) — the core concurrency primitive used here.
- [`ZIO.timeout` / `timeoutFail`](https://zio.dev/reference/core/zio/#timeout-operations) — for fail-fast behavior.
- [`Ref`](https://zio.dev/reference/concurrency/ref/) — useful if you want to track/expose current concurrency as a metric.

---

## How client + server work together

```
Agent (this PR)                          Server (proposed complement)
─────────────────                        ────────────────────────────
Request fails (5xx/IOException)          Too many in-flight queries
        │                                        │
        ▼                                        ▼
Retry with exponential backoff           Semaphore rejects with 503
  + jitter (spread retries)                (fail-fast, not queue)
        │                                        │
        ▼                                        ▼
Server has time to recover               Pool connections free up
        │                                        │
        └──────── Next retry succeeds ───────────┘
```

**The two mechanisms are complementary:**
- The **server** protects itself (sheds load it can't handle).
- The **agent** handles the resulting 503s gracefully (retries with backoff).
- Together they create a **stable feedback loop** instead of a cascading failure.

---

## Bottom line

✅ **This PR is a solid step** — merge-worthy on its own merits.

🔧 **Follow-up recommendation**: add a server-side `Semaphore` in `TransactorProvider` (single-line change) to cap concurrent DB access. Without it, the retry mechanism can amplify load during sustained outages. A proof-of-concept test demonstrating this is available at `server/src/test/.../ThrottledCheckpointServiceDemoTests.scala` in this branch.

agent/src/main/resources/reference.conf
lsulak: wow, thanks for the review & doing it so quickly even! This is used here:

```scala
    val okHttpClient = new OkHttpClient.Builder()
      .connectTimeout(connectTimeoutMs, TimeUnit.MILLISECONDS)
      .readTimeout(readTimeoutMs, TimeUnit.MILLISECONDS)
      .writeTimeout(writeTimeoutMs, TimeUnit.MILLISECONDS)
      .build()
```

so these values are used in the OkHTTP meaning that it's all on the HTTP layer: https://www.baeldung.com/okhttp-timeouts

lsulak: From the source above:

ConnectionTimeout: It defines a time period in which our client should establish a connection with a target host.

ReadTimeout: It defines a maximum time of inactivity between two data packets when waiting for the server’s response.

WriteTimeout: it defines a maximum time of inactivity between two data packets when sending the request to the server.

agent/src/main/resources/reference.conf
lsulak: I would tend to go more defensive on this one. 

In healthy stable environment it might be that there are almost no retries or 1 is enough, but there were some production issues and I would like the agent to be generally / by default a bit more robust. 

So, basically if there are 3 potential retries and it's not used in healthy environment, there is zero cost to have configured 3. But under the load during peaks / different environments this can be beneficial.

